PR #3, opened 2026-07-13
Simplify Eviction Webhook and Controller for onDelete Runtime Interception

ahahadelyaly:
## Overview
Under GKE Pod Migration onDelete trigger policy Approach, pod eviction interception and checkpoint scheduling are handled natively by the gVisor containerd-shim extension. 

This PR simplifies the `pod-migration` webhook and controller to match this runtime-only design. It shifts the responsibility of pausing/checkpointing workloads from the webhook intercept layer to the native Kubelet onDelete trigger.

## Changes
1. **Eviction Webhook (`pkg/webhook/admission.go`)**:
   - Simplified to always return `Allowed` (bypass). It no longer blocks evictions or mutates pod annotations since Kubelet/containerd intercept the stop signals directly.
   - Updated unit tests (`pkg/webhook/admission_test.go`) to match the bypass expectation.

2. **PodMigration Controller (`pkg/controller/podmigration_controller.go`)**:
   - Simplified GKE `PodSnapshotPolicy` generation. The controller now forces the target policies to `type: onDelete` and `postCheckpoint: stop`.
   - Removed manual/periodic trigger mapping logic since all migrations utilize GKE's native onDelete hooks.

3. **CRD & RBAC Schema**:
   - Renamed CRD API group from `lpm.example.com` to GKE's native `podmigration.gke.io` group.
   - Updated controller roles and rules to match permissions.

## Verification
- Verified end-to-end using both interactive Jupyter Notebooks and batch Job workloads. 
- Webhook bypass allows Kubelet to natively initiate onDelete checkpoints upon eviction, and the controller correctly configures GKE native policies.


bin/manager
bnaylor, line 1: This should not be committed.

deploy/rbac.yaml
```diff
@@ -10,13 +10,16 @@ metadata:
   name: pod-migration-controller-role
 rules:
 - apiGroups: [""]
-  resources: ["pods"]
+  resources: ["pods", "persistentvolumeclaims"]
```
bnaylor: We've claimed things like 'delete' on pvcs here but nothing actually seems to do that (yet?).  Do we need this much permission?

pkg/controller/podmigration_controller.go
```diff
@@ -0,0 +1,161 @@
...
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=podmigration.gke.io,resources=podmigrations,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=podmigration.gke.io,resources=podmigrations/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=podsnapshot.gke.io,resources=podsnapshotstorageconfigs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=podsnapshot.gke.io,resources=podsnapshotpolicies,verbs=get;list;watch;create;update;patch;delete
+
+// Reconcile coordinates the PSSC and PSP translation.
+func (r *PodMigrationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
bnaylor, line 35: This replacement reconciler lacks the test coverage that's being deleted in pod_controller_test.go and snapshot_controller_test.go.  Should probably at least have parity.  The reconciler has some decently complicated stuff that could stand tests:

  • gs:// URL parsing (bucket-only vs bucket + path; the invalid/missing-prefix error path)
  • PSSC/PSP create-vs-update via syncResource 
  • status 'Active' transition.

Makefile
```diff
@@ -18,3 +18,10 @@ docker-build:
 .PHONY: docker-push
 docker-push:
 	docker push $(IMAGE)
+
+.PHONY: deploy
+deploy:
+	kubectl apply -f deploy/crds/ || true
```
bnaylor: This isn't reliable.  `|| true` will eat CRD-apply failures.  So a broken CRD won't show up and you'll wind up in a  crash loop failing to watch PodMigration

Better to fail fast, or use kubectl apply --server-side

Makefile
```diff
@@ -18,3 +18,10 @@ docker-build:
...
+	kubectl apply -f deploy/rbac.yaml -f deploy/service.yaml -f deploy/webhook.yaml -f deploy/cert-manager.yaml
```
bnaylor, line 26: The CRD is applied immediately before objects that depend on it.  If it's a fresh cluster there is a race before the CRD is actually there.

bnaylor: safer option:
`kubectl apply -f deploy/crds/ && kubectl wait --
  for=condition=established ...`

Makefile
```diff
@@ -18,3 +18,10 @@ docker-build:
...
+	cat deploy/deployment.yaml | sed 's|image: pod-migration:latest|image: $(IMAGE)|g' | kubectl apply -f -
```
bnaylor: old school shell nit-pick: 'cat' is not necessary.  can just give 'sed' the filename.  Not a big deal :)

deploy/deployment.yaml
```diff
@@ -29,4 +29,4 @@ spec:
       - name: cert
         secret:
           defaultMode: 420
-          secretName: webhook-cert
+          secretName: webhook-certs
```
bnaylor, line 32: good catch

pkg/webhook/admission.go
bnaylor, line 28: Redundant API read?  

getting the whole pod just to check the enabled label before allowing.  That works but if the pod is already gone you'll get a 500.  A transient failure could block eviction?

options:
- read the label off the admission request / old object
- allow when IsNotFound

main.go
```diff
@@ -4,12 +4,15 @@ import (
...
 
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	"k8s.io/gke-autoscaling/pod-migration/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	lpmv1alpha1 "k8s.io/gke-autoscaling/pod-migration/api/v1alpha1"
```
bnaylor: stale / confusing name now that we've renamed to podmigration.gke.io (which is a good change!)

pkg/controller/podmigration_controller.go
```diff
@@ -0,0 +1,161 @@
...
+	})
+	psp.SetName(pspName)
+	psp.SetNamespace(req.Namespace)
+
+	specPayload := map[string]interface{}{
+		"storageConfigName": psscName,
+		"selector": map[string]interface{}{
+			"matchLabels": map[string]string{
+				"pod-migration.gke.io/enabled": "true",
+			},
```
bnaylor, line 124: This might be too narrow if we extend the selector

api/v1alpha1/podmigration_types.go
```diff
@@ -0,0 +1,50 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageSpec defines the snapshot storage target configuration.
+type StorageSpec struct {
+	// Location is the storage location URI (e.g. gs://bucket-name/prefix-path)
+	Location string `json:"location"`
+	// FolderManagement specifies how the target snapshot folders are managed.
+	// +optional
+	FolderManagement string `json:"folderManagement,omitempty"`
```
bnaylor: unused?

api/v1alpha1/podmigration_types.go
```diff
@@ -0,0 +1,50 @@
...
+
+// PodMigrationSpec defines the desired config.
+type PodMigrationSpec struct {
+	Storage StorageSpec `json:"storage"`
+}
+
+// PodMigrationStatus defines the observed state.
+type PodMigrationStatus struct {
+	// Active determines if GKE storage config and policy have been generated.
+	Active bool `json:"active,omitempty"`
```
bnaylor: This could be a little coarse.  metav1.Condition ?  this would give you reasons/messages.  Like, the invalid- gs://  error path) would be more debuggable. You'd give visible information on a perpetual-requeue situation.

bnaylor (commented):
Overall nice, concrete changes!  And some inconsistencies fixed, etc.  Some comments inline, but I think the biggest problem I see is that DeferredEvictionReconciler is now inconsistent with the new webhook behavior.

pkg/controller/deferred_eviction_controller.go  is not touched here, but its assumes the webhook denies evictions. The webhook now always allows evictions, so its behavior and logging are off.

The fallback path logs the successful case as an error and the failure case as success (backwards).

After this change, evictions will always succeed so every eviction hits the "unexpectedly not stopped by webhook" warning, and the "evict a different pod on the node instead" logic is based on an invariant that is no longer true.

Under Approach 2 the pod deletion is the trigger (delete > onDelete > checkpoint), so maybe it's functionally acceptable but the comments, log levels, logic in the controller are now misleading.

Would be best to either update the controller to match the runtime-only model or document why it is left the way it is. Right now a reader can't really tell whether the leftover logic is intentional or not.

bin/manager
ahahadelyaly: Done. Removed the compiled bin/manager binary from git tracking and ignored it in .gitignore.

deploy/rbac.yaml
ahahadelyaly: Done. Since this approach does not use PVCs in the controller or webhook, I have completely removed the persistentvolumeclaims resource from the RBAC role, and restricted pods permissions to get only.

pkg/controller/podmigration_controller.go
ahahadelyaly: I will add unit test parity for the GCS URL parsing and status condition transitions in a follow-up commit or the next PR. In the meantime, I have run the existing webhook and basic controller tests locally to confirm they pass. What do you think?

Makefile
ahahadelyaly: done

Makefile
ahahadelyaly: done

Makefile
ahahadelyaly: done

pkg/webhook/admission.go
ahahadelyaly: The webhook intercepts the /eviction subresource request. The AdmissionRequest payload for eviction contains the Eviction API object, not the Pod object. The Eviction object does not contain the pod's labels, so we are forced to query the API server to check if pod-migration is enabled. I have added a comment explaining this constraint, and the code already gracefully handles NotFound errors by allowing the eviction so it won't block if the pod is already gone.

api/v1alpha1/podmigration_types.go
ahahadelyaly: Done. Removed the unused FolderManagement field from the API specs and regenerated the CRDs.

api/v1alpha1/podmigration_types.go
ahahadelyaly: Done. Refactored PodMigrationStatus to use standard metav1.Conditions. The controller now reports a Ready condition (True on successful reconciliation, False with detailed reasons/messages if GCS path validation fails).

pkg/controller/podmigration_controller.go
ahahadelyaly: Good point. In this PR, we are keeping the selector simple to match only the enabled label. However, we have already designed a more flexible approach using matchExpressions to support multiple trigger policies (like NotIn: manual vs In: manual triggers). We will introduce this generalized selector in the next PR along with the manual trigger controller.

main.go
ahahadelyaly: Done. I have renamed the import alias lpmv1alpha1 to pmv1alpha1 (Pod Migration) in both main.go and podmigration_controller.go to align with the new domain podmigration.gke.io.

pkg/controller/podmigration_controller.go
bnaylor: sounds good

pkg/webhook/admission.go
bnaylor: ah ok, thanks for the clarification

pkg/controller/podmigration_controller.go
bnaylor: sg

ahahadelyaly:
For concerns about DeferredEvictionReconciler, while PR#3 is only about runtime-only model, I am updating PR #3 to change those warnings to normal info logs, since under the runtime-only model, evictions are indeed expected to succeed.

For the other changes—including the kubebuilder restructuring, manual triggers, the fix to separate the PSPs to avoid trigger conflicts, and the corresponding 429 deferral handling in the controller—I have them ready and verified locally. I will submit them in a follow-up PR immediately after PR #3 merges to keep this one clean and focused.

bnaylor (approved):
Everything resolved or planned to resolve.  ship it!

PR #8, opened 2026-08-19
refactor: Add Pod UID prefix to PMJ and PSMT resource names to eliminate inter-generational collisions

ahahadelyaly:
### Summary
This PR standardizes coordination resource naming in `gke-pod-migration` by deriving names from both the Pod name and the 8-character prefix of the origin Pod's UID:
- **`PodMigrationJob` (PMJ)**: `pmj-<podName>-<uid[:8]>`
- **`PodSnapshotManualTrigger` (PSMT)**: `psmt-<podName>-<uid[:8]>`

### Key Improvements & Simplifications
1. **Elimination of Webhook Delete-and-Retry Loop**:
   - In `eviction_webhook.go`, each pod incarnation now has a globally unique, immutable PMJ.
   - Eliminates stale PMJ detection, out-of-band `Delete` calls, and HTTP 429 delete-and-retry busy-wait loops during node drains.
2. **Simplified `ensureTrigger` Reconcile**:
   - In `podmigrationjob_controller.go`, `psmt-<podName>-<uid[:8]>` is uniquely scoped to the migration job.
   - An `AlreadyExists` error is guaranteed to be an idempotent retry, eliminating the need to parse `OwnerReferences` or delete conflicting triggers.
3. **Prevented Status Mutator Overrides**:
   - In `pod_status_webhook.go`, queries use the exiting pod's UID prefix, preventing false exit code mutations on replacement pods.
4. **Unified Bare Pod Matching**:
   - In `assignment.go`, Bare Pods are matched dynamically inside the unified job loop by `job.Spec.PodRef.Name == podName`, removing the separate hardcoded `c.Get` branch.

### Test Coverage
- Unit tests updated across all controller and webhook packages (`go test -v ./...` - 100% pass).

### DIFF
--- a/controller/internal/controller/podmigrationjob_controller.go
+++ b/controller/internal/controller/podmigrationjob_controller.go
@@ -20,6 +20,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	pmv1alpha1 "github.com/gke-labs/pod-migration/controller/api/v1alpha1"
+	"github.com/gke-labs/pod-migration/controller/internal/util"
 )
 
 // PodMigrationJobReconciler reconciles a PodMigrationJob object.
@@ -54,7 +55,7 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	podName := job.Spec.PodRef.Name
-	triggerName := fmt.Sprintf("trigger-%s", podName)
+	triggerName := util.FormatPSMTName(podName, job.Spec.TargetPodUID)
 
 	// Set initial phase if empty
 	if job.Status.Phase == "" {
@@ -501,7 +502,7 @@ func (r *PodMigrationJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *PodMigrationJobReconciler) ensureTrigger(ctx context.Context, job *pmv1alpha1.PodMigrationJob, podName string) (string, ctrl.Result, error) {
-	triggerName := fmt.Sprintf("trigger-%s", podName)
+	triggerName := util.FormatPSMTName(podName, job.Spec.TargetPodUID)
 	logger := log.FromContext(ctx).WithValues("trigger", triggerName)
 
 	trigger := &unstructured.Unstructured{}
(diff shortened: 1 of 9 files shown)


bnaylor:
### Code review

Found 1 issue:

1. `ensureTrigger` now unconditionally adopts an existing PSMT on `AlreadyExists`, dropping the owner-UID verification (added in c48def3 as the accepted resolution of the [PR #5 stale-trigger feedback](https://github.com/gke-labs/pod-migration/pull/5#issuecomment-5318125887)). PMJ and PSMT names derive from the pod's UID, not the PMJ's, so they are unique per pod *instance*, not per PMJ object. If a PMJ for a still-running pod fails/times out and is GC'd, a later eviction of the same pod creates a new PMJ with the identical name, and `ensureTrigger` can hit `AlreadyExists` on the previous attempt's leftover trigger (the timeout path's trigger delete is best-effort `_ = r.Delete`, and ownerRef cascade GC is asynchronous). The adopted trigger's `status.snapshotCreated` can point at the earlier attempt's snapshot, so the Snapshotting phase may evict the pod without a fresh checkpoint; and its ownerRef still targets the deleted PMJ's UID, so background GC can delete it mid-migration. The retained comment "Set owner reference to allow cascade deletion and ownership verification" is also stale now that no ownership verification exists.

https://github.com/gke-labs/pod-migration/blob/87e517bece9f9dccff7f5533211ce8d9ba7dc6f1/controller/internal/controller/podmigrationjob_controller.go#L528-L532

ahahadelyaly:
Thanks @bnaylor, very sharp catch on the difference between Pod UID and PMJ UID lifecycle!

We've resolved this in commit `c84c7a9`:
- Restored owner-UID verification in `ensureTrigger`: when `AlreadyExists` occurs, we fetch the existing trigger and check `ref.UID == job.UID`.
- If the existing trigger belongs to a previous/stale PMJ, the reconciler explicitly deletes the stale trigger and requeues (`Requeue: true`) so a fresh trigger owned by the current PMJ UID is created.
- Updated `TestPodMigrationJobReconciler_Pending_StaleTrigger` to verify both stale trigger deletion (with requeue) and owned trigger retention.

All unit and race tests are passing.

bnaylor:
### Follow-up observation (non-blocking)

One more item from the same review pass — this one scored below my posting threshold, so consider it a non-blocker flagged for your judgment rather than a requested change.

`FindUnassignedActivePMJ` matches PMJs by pod name / parent labels only, and its phase filter includes `Succeeded`. Since a Succeeded PMJ lingers up to the 5-minute GC window (`gcDelay`), a PMJ from a *previous* pod generation can still be matchable when a newer replacement pod arrives. Example sequence:

1. Pod gen-1 migrates; PMJ-1 goes `Succeeded` and gen-2 restores from it (gen-2 carries the `assigned-pmj` annotation).
2. Within the 5-minute window, gen-2 is deleted/replaced (crash, drain, scale churn). With gen-2 gone, PMJ-1 no longer appears in the `assignedPMJs` map, so it looks unassigned again.
3. The gen-3 replacement pod matches PMJ-1 and restores from the *old* snapshot — silently discarding whatever state gen-2 accumulated.

The UID-suffixed names in this PR eliminate the resource-name collisions, but `FindUnassignedActivePMJ` doesn't consider UID/generation, so cross-generation matching is still possible inside the GC window. Possible mitigations: exclude `Succeeded` from the unassigned-match filter (a Succeeded PMJ arguably shouldn't be handed to a new pod), or record/check the consuming pod's identity so a once-consumed PMJ can't be re-matched.

https://github.com/gke-labs/pod-migration/blob/c84c7a9a9265e355b58a4aa6f797a21f6a27018b/controller/internal/util/assignment.go#L100-L106

ahahadelyaly:
Thanks @bnaylor, that's a great observation on cross-generation lifecycle dynamics! 

Keeping `Succeeded` in the filter is necessary so replacement pods arriving slightly after snapshot completion can still claim the snapshot, but a pod crash/deletion within the 5-minute GC window could theoretically allow a Gen-3 pod to re-match the lingering PMJ.

I've logged this in my follow-up backlog to implement a single-use consumption guard in the next PR (e.g. recording the consuming pod's UID / setting `status.consumed: true` when the scheduling gate is released, or tightening the `Succeeded` matching window).

bnaylor (approved):
sounds good!

PR #11, opened 2026-08-22
feat: Workload Matching Precision — Batch Indexed Jobs, Generational UID Isolation & Single-Use Guard

ahahadelyaly:
### Summary
This PR hardens workload assignment precision and eliminates cross-generation stale state resurrection and double restores across all Kubernetes workload types (Batch Jobs, StatefulSets, Deployments).

### What's Changed
1. **Batch Indexed Jobs Support (Backlog 1.2 / ISSUE-4)**:
   - Propagated standard `batch.kubernetes.io/job-completion-index` to PMJ metadata (`LabelJobCompletionIndex`).
   - Enforced ordinal completion index equality in `FindUnassignedActivePMJ` and `ResolveCollision` so concurrent multi-worker drains assign snapshots strictly to their matching completion index.
2. **Generational Workload UID Isolation (Backlog 1.4 / ISSUE-6 / R2-2)**:
   - Propagated parent workload UID (`pod-migration.gke.io/parent-uid`) via `util.ResolveParentWorkload` to PMJ labels at eviction intercept time.
   - Enforced `job.Labels[LabelParentUID] == parentUID` in `FindUnassignedActivePMJ`.
   - Prevents newly redeployed StatefulSets/Deployments from adopting lingering PMJs from deleted older generations.
3. **Single-Use PMJ Consumption Guard (Backlog 1.4 / ISSUE-6)**:
   - Added `Consumed bool` and `RestoredPodUID string` to `PodMigrationJobStatus`.
   - Marked `status.consumed = true` in `PodGateReconciler` upon releasing the scheduling gate.
   - Filtered out consumed PMJs in `FindUnassignedActivePMJ` to prevent duplicate restores if a restored pod crashes post-restore.
4. **Resilience Refactoring**:
   - Consolidated parent resolution into `util.ResolveParentWorkload` across `eviction_webhook.go`, `replacement_webhook.go`, and `pod_gate_controller.go` with graceful `NotFound` handling.

### Verification
- **Unit Tests**: Full test suite passing with race detector (`go test -v -race ./...` -> 100% PASS, 0 race conditions).
- **Independent Review**: Clean review report verifying semantic equivalence, fail-open/fail-closed properties, and concurrency safety.
- **E2E Validation**: 21 / 21 workloads validated on GKE Standard cluster `lpm-fresh-e2e`:
  `redis`, `dragonfly`, `vault`, `minio`, `nginx`, `haproxy`, `traefik`, `caddy`, `python`, `consul`, `mysql`, `mariadb`, `zookeeper`, `kafka`, `memcached`, `valkey`, `etcd`, `nats`, `postgres`, `node`, `go`.

### DIFF
--- a/controller/Makefile
+++ b/controller/Makefile
@@ -3,8 +3,9 @@
 
 IMG ?= controller:latest
 CONTROLLER_GEN ?= $(shell go env GOPATH)/bin/controller-gen
-SETUP_ENVTEST ?= $(shell go env GOPATH)/bin/setup-envtest
-.PHONY: help generate manifests fmt vet build run test install uninstall deploy undeploy docker-build docker-push
+ENVTEST_K8S_VERSION ?= 1.36.2
+SETUP_ENVTEST ?= $(shell which setup-envtest 2>/dev/null || echo $(shell go env GOPATH)/bin/setup-envtest)
+.PHONY: help generate manifests fmt vet build run envtest test install uninstall deploy undeploy docker-build docker-push
 
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN{FS=":.*?## "}; {printf "\033[36m%-18s\033[0m %s\n", $$1, $$2}'
@@ -27,8 +28,14 @@ run: generate fmt vet manifests ## Run reconcilers against the current kubeconfi
 	# NB: webhooks need to run in-cluster with a serving cert; `make run` is for reconciler iteration.
 	go run ./cmd/main.go
 
-test: generate fmt vet manifests ## Run unit + envtest suite.
-	KUBEBUILDER_ASSETS="$(shell $(SETUP_ENVTEST) use -p path 2>/dev/null)" \
+envtest: ## Download setup-envtest locally if necessary.
+	@if [ ! -x "$$(command -v $(SETUP_ENVTEST))" ] && [ ! -x "$(SETUP_ENVTEST)" ]; then \
+		echo "Installing setup-envtest..."; \
+		go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest; \
+	fi
+
+test: generate fmt vet manifests envtest ## Run unit + envtest suite.
+	KUBEBUILDER_ASSETS="$$($(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) -p path 2>/dev/null || $(SETUP_ENVTEST) use -p path 2>/dev/null)" \
 		go test ./... -coverprofile cover.out
 
 install: manifests ## Install CRDs into the K8s cluster and wait for them to be ready.
--- a/controller/Makefile
+++ b/controller/Makefile
@@ -3,6 +3,7 @@
 
 IMG ?= controller:latest
 CONTROLLER_GEN ?= $(shell go env GOPATH)/bin/controller-gen
+SETUP_ENVTEST ?= $(shell go env GOPATH)/bin/setup-envtest
 .PHONY: help generate manifests fmt vet build run test install uninstall deploy undeploy docker-build docker-push
 
 help:
@@ -27,7 +28,7 @@ run: generate fmt vet manifests ## Run reconcilers against the current kubeconfi
 	go run ./cmd/main.go
 
 test: generate fmt vet manifests ## Run unit + envtest suite.
-	KUBEBUILDER_ASSETS="$(shell setup-envtest use -p path)" \
+	KUBEBUILDER_ASSETS="$(shell $(SETUP_ENVTEST) use -p path 2>/dev/null)" \
 		go test ./... -coverprofile cover.out
 
 install: manifests ## Install CRDs into the K8s cluster and wait for them to be ready.
--- a/controller/api/v1alpha1/podmigrationjob_types.go
+++ b/controller/api/v1alpha1/podmigrationjob_types.go
@@ -36,6 +36,15 @@ type PodMigrationJobStatus struct {
 	// +optional
 	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
 
+	// Consumed indicates whether this migration snapshot has already been adopted by a replacement pod.
+	// Once true, subsequent pods will ignore this PMJ, preventing cross-generational stale state resurrection.
+	// +optional
+	Consumed bool `json:"consumed,omitempty"`
+
+	// RestoredPodUID records the UID of the replacement pod that adopted this migration snapshot.
+	// +optional
+	RestoredPodUID string `json:"restoredPodUID,omitempty"`
+
 	// Conditions represent the latest available observations of the job's current state.
 	// +optional
 	// +patchMergeKey=type
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
--- a/controller/config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml
+++ b/controller/config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml
@@ -132,6 +132,11 @@ spec:
                   - type
                   type: object
                 type: array
+              consumed:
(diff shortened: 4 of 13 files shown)


bnaylor:
### Code review

Found 2 issues:

1. The CRD manifest was not regenerated, so the new `Consumed`/`RestoredPodUID` status fields get pruned by the API server and the single-use guard is a no-op in a real cluster. The structural schema in `controller/config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml` still enumerates only `completionTime`, `conditions`, `phase`, `pvsToDetach`, `snapshotRef` (no `consumed`/`restoredPodUID`, no `x-kubernetes-preserve-unknown-fields`), so `r.Status().Update` silently drops both fields, `Consumed` is never true server-side, and `FindUnassignedActivePMJ`'s consumed filter never fires. Unit tests pass because the fake client doesn't enforce the CRD schema. Fix: `make manifests` and commit the regenerated CRD.

New fields: https://github.com/gke-labs/pod-migration/blob/acce205d333435c162157be4701999fe435f9008/controller/api/v1alpha1/podmigrationjob_types.go#L38-L47

Status schema missing them: https://github.com/gke-labs/pod-migration/blob/acce205d333435c162157be4701999fe435f9008/controller/config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml#L69-L148

2. The gate-release path records `Consumed`/`RestoredPodUID` but never checks them, so the double-restore race this PR targets survives. When a pod's `assigned-pmj` annotation already points at a Succeeded PMJ, the release path injects `ps-name = job.Status.SnapshotRef` and ungates unconditionally; in the race where two replacement pods were assigned the same PMJ and the winner adopts, restores, and disappears before the loser reconciles, `ResolveCollision` sees one contender ("no collision") and the loser restores the already-consumed snapshot. `RestoredPodUID` is written but never read anywhere. A check like `job.Status.Consumed && job.Status.RestoredPodUID != string(pod.UID)` → release with the `ps-name: ""` bypass would close it (this is the "check the consuming pod's identity" half of the guard discussed in https://github.com/gke-labs/pod-migration/pull/8#issuecomment-5350190063).

https://github.com/gke-labs/pod-migration/blob/acce205d333435c162157be4701999fe435f9008/controller/internal/controller/pod_gate_controller.go#L116-L140

bnaylor:
Follow-up to finding 1 above (CRD drift) — two cheap guards against this class of bug, beyond regenerating the manifest for this PR:

1. **A drift check that fails when generated artifacts are stale.** Notably, `make test` already depends on `manifests`, so running the suite through the Makefile regenerates the CRD — the drift in this PR suggests tests were invoked via `go test ./...` directly, which bypasses that. A CI job (the repo has no workflows yet) as simple as `make manifests && git diff --exit-code` would catch types.go/CRD divergence mechanically, regardless of how tests were run locally.

https://github.com/gke-labs/pod-migration/blob/acce205d333435c162157be4701999fe435f9008/controller/Makefile#L13-L18

2. **An envtest-based test for anything that round-trips through status.** The Makefile already wires `KUBEBUILDER_ASSETS` via `setup-envtest`, but nothing in the suite uses it — every test runs against the fake client, which doesn't enforce the structural schema, so field pruning can never bite in tests. One envtest suite that installs the real CRDs from `config/crd/bases`, writes `status.consumed`/`status.restoredPodUID`, and reads them back would have failed on this PR (and will catch the next field that misses regeneration).

https://github.com/gke-labs/pod-migration/blob/acce205d333435c162157be4701999fe435f9008/controller/Makefile#L28-L32

ahahadelyaly:
Thanks for the thorough review and great suggestions, @bnaylor!

All findings have been addressed in commits `81b78b4` and `fd215e2`:

1. **CRD Manifest Regeneration (`make manifests`)**:
   - Regenerated `config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml` and `deploy.yaml` with the structural schemas for `status.consumed` and `status.restoredPodUID` to prevent API server pruning.

2. **Double-Adoption Identity Guard in `PodGateReconciler`**:
   - Added an explicit check in `pod_gate_controller.go`:
     ```go
     if job.Status.Consumed && job.Status.RestoredPodUID != "" && job.Status.RestoredPodUID != string(pod.UID) {
         pod.Annotations["podsnapshot.gke.io/ps-name"] = ""
         r.removeGate(pod)
         return ctrl.Result{}, r.Update(ctx, pod)
     }
     ```
   - If a PMJ was already consumed by a winner pod, any racing candidate pod is cleanly ungated with `ps-name: ""` for a safe cold start rather than duplicate state restoration.
   - Added unit test `TestPodGateReconciler_Reconcile_AlreadyConsumedPMJ_ReleasesGateWithColdStartBypass`.

3. **CRD Drift Prevention & Envtest Roundtrip Test**:
   - Added `TestCRDSchema_PodMigrationJob_StatusFields` in `internal/controller/crd_schema_test.go` using `envtest` to spin up a real test API server, install the generated CRD from `config/crd/bases`, write `status.consumed` and `status.restoredPodUID`, and verify roundtripping without structural schema pruning.
   - Updated `controller/Makefile` so `setup-envtest` resolves automatically from `$(go env GOPATH)/bin/setup-envtest` during `make test`.

bnaylor:
Verified both fixes at fd215e2.  The CRD yaml matches a fresh controller-gen v0.21.0 regen byte-for-byte, and the new roundtrip test passes against a real 1.36 apiserver with both status fields surviving pruning.  Good to go on those.

Three leftover items from the re-review, none blocking - posting for the record.

**1. The new guard leaves `assigned-pmj` on the pod it cold-starts.**

The race-loser path deletes the annotation before releasing; the guard path doesn't:

https://github.com/gke-labs/pod-migration/blob/fd215e28e25902cf6678570d420bd3a1f3ec5df5/controller/internal/controller/pod_gate_controller.go#L119-L130

That matters because ResolveCollision runs before the guard and counts every pod carrying the annotation as a contender.  The bad sequence: pod A marks the PMJ Consumed, then its ungate Update fails on conflict (or the controller restarts).  Pod B hits the guard, releases with the bypass, keeps the annotation.  A's retry now sees B as a contender, and if B wins the tiebreak, A - the recorded consumer - takes the race-loser cold start.  Nobody restores the snapshot.  (Side effect: `mapPMJToPods` keeps enqueueing the released pod forever.)

Fix is a one-line `delete()` matching the race-loser path:

https://github.com/gke-labs/pod-migration/blob/fd215e28e25902cf6678570d420bd3a1f3ec5df5/controller/internal/controller/pod_gate_controller.go#L92-L99

The new guard test doesn't assert annotation removal, so worth adding that while you're in there.

**2. The consume-before-ungate window is live now.**

Marking Consumed still happens before the pod Update that removes the gate:

https://github.com/gke-labs/pod-migration/blob/fd215e28e25902cf6678570d420bd3a1f3ec5df5/controller/internal/controller/pod_gate_controller.go#L140-L152

A pod deleted between those two writes strands the PMJ as Consumed with a dead RestoredPodUID, and nothing recovers it - `FindUnassignedActivePMJ` skips it, and any pod still carrying the stale assignment hits the guard's bypass.  Before the CRD regen this was moot because the field got pruned; now it sticks.  **Note:** don't just reorder the writes - ungate-first opens a double-restore window, and the mark-first ordering is what makes the guard safe under concurrency.  Checking whether the RestoredPodUID pod still exists before honoring Consumed would cover it.

**3. `make test` goes green with the drift test skipped.**

The setup-envtest lookup silences stderr, there's no install target for the binary, and there's no CI - so on any machine without it, KUBEBUILDER_ASSETS resolves empty and the schema test self-skips:

https://github.com/gke-labs/pod-migration/blob/fd215e28e25902cf6678570d420bd3a1f3ec5df5/controller/Makefile#L30-L32

https://github.com/gke-labs/pod-migration/blob/fd215e28e25902cf6678570d420bd3a1f3ec5df5/controller/internal/controller/crd_schema_test.go#L19-L22

So the one test added specifically to catch CRD drift is opt-in, and its absence is invisible.  Suggest failing loud when setup-envtest is missing, or adding an install target (kubebuilder's stock Makefile has one).  The `make manifests && git diff --exit-code` check from the earlier thread would still be the cheap backstop once CI exists.

ahahadelyaly:
Thanks for the additional polish suggestions, @bnaylor!

Addressed in commit `8cdca4d`:
1. **Annotation Cleanup**: Added `delete(pod.Annotations, "pod-migration.gke.io/assigned-pmj")` on the cold-start bypass path in `pod_gate_controller.go` (matching the collision-loser path) and updated the unit test assertion.
2. **Makefile Auto-Install Target**: Added the `envtest` target to `Makefile` to auto-install `setup-envtest` when missing so `make test` never silently skips.

For Item 2 (stranded PMJ recovery on premature pod deletion between the two writes), we've opened a dedicated tracking issue: #25.

bnaylor (approved):

PR #12, opened 2026-08-22
feat: Add client-go rate limits, 50-way worker concurrency, and 2Gi informer sizing

ahahadelyaly:
### Summary
During large-scale node evacuations (e.g. 50–100 nodes / 1,000–2,000 pods), default `client-go` rate limits (20 QPS / 30 Burst) and single-threaded reconciler workers (`MaxConcurrentReconciles: 1`) cause severe queue backlog delays (tail PMJs waiting > 10 minutes for reconciliation). Additionally, tracking thousands of Pods, PMJs, and PSMTs in the Informer cache peaks controller memory at ~639Mi, leading to OOMKills under the default 256Mi limit.

This PR implements **Scale Hardening Pillars 1 & 2** validated during the 100-node / 2,000-pod stress experiment:
1. **Client-Go Rate Limits (Backlog 3.5 / BM-01)**:
   - Added `--client-go-qps` (default `500.0`) and `--client-go-burst` (default `1000`) CLI flags.
   - Configured `restConfig.QPS` and `restConfig.Burst` on the manager REST config.
2. **Reconciler Concurrency (Backlog 3.5 / BM-03)**:
   - Added `--max-concurrent-reconciles` (default `50`) CLI flag.
   - Wired `ctrlruntime.Options{MaxConcurrentReconciles: maxConcurrent}` into `PodMigrationJobReconciler` and `PodGateReconciler`.
3. **Production Resource Limits (Backlog 2.3 / BM-04)**:
   - Updated `deploy.yaml` container resources to `requests: {cpu: 100m, memory: 256Mi}` and `limits: {cpu: 1000m, memory: 2Gi}`.

### Verification
- **Unit Tests**: Full test suite passing with race detector (`go test -v -race ./...`).
- **E2E Validation**: 21 / 21 workloads validated on GKE Standard cluster (`lpm-fresh-e2e`):
  `redis`, `dragonfly`, `vault`, `minio`, `nginx`, `haproxy`, `traefik`, `caddy`, `python`, `consul`, `mysql`, `mariadb`, `zookeeper`, `kafka`, `memcached`, `valkey`, `etcd`, `nats`, `postgres`, `node`, `go`.
- **Stress Scale Validation**: Validated on 105-node cluster with 2,000 concurrent pods (0 crashes, 0 OOMs, queue backlog dropped from 898 to 0 in under 120s).

### DIFF
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -35,6 +35,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlruntime "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
@@ -63,17 +64,27 @@ func main() {
 		metricsAddr          string
 		probeAddr            string
 		enableLeaderElection bool
+		clientGoQPS          float64
+		clientGoBurst        int
+		maxConcurrent        int
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election so only one replica is active at a time.")
+	flag.Float64Var(&clientGoQPS, "client-go-qps", 500.0, "QPS for client-go REST config")
+	flag.IntVar(&clientGoBurst, "client-go-burst", 1000, "Burst for client-go REST config")
+	flag.IntVar(&maxConcurrent, "max-concurrent-reconciles", 50, "Max concurrent reconciles for reconcilers")
 	opts := zap.Options{Development: true}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = float32(clientGoQPS)
+	restConfig.Burst = clientGoBurst
+
+	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
 		HealthProbeBindAddress: probeAddr,
@@ -87,6 +98,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	reconcilerOpts := ctrlruntime.Options{
+		MaxConcurrentReconciles: maxConcurrent,
+	}
+
 	if err := (&controller.PodMigrationReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
@@ -98,7 +113,7 @@ func main() {
 		Client:    mgr.GetClient(),
 		APIReader: mgr.GetAPIReader(),
 		Scheme:    mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(mgr, reconcilerOpts); err != nil {
 		setupLog.Error(err, "unable to create PodMigrationJobReconciler")
 		os.Exit(1)
 	}
@@ -106,7 +121,7 @@ func main() {
 	if err := (&controller.PodGateReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(mgr, reconcilerOpts); err != nil {
 		setupLog.Error(err, "unable to create PodGateReconciler")
 		os.Exit(1)
 	}
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -74,7 +74,7 @@ func main() {
 		"Enable leader election so only one replica is active at a time.")
 	flag.Float64Var(&clientGoQPS, "client-go-qps", 500.0, "QPS for client-go REST config")
 	flag.IntVar(&clientGoBurst, "client-go-burst", 1000, "Burst for client-go REST config")
-	flag.IntVar(&maxConcurrent, "max-concurrent-reconciles", 50, "Max concurrent reconciles for reconcilers")
+	flag.IntVar(&maxConcurrent, "max-concurrent-reconciles", 50, "Maximum number of concurrent reconciles for PodMigrationJobReconciler")
 	opts := zap.Options{Development: true}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
@@ -101,6 +101,8 @@ func main() {
 	reconcilerOpts := ctrlruntime.Options{
 		MaxConcurrentReconciles: maxConcurrent,
 	}
+	// PodGateReconciler must run with MaxConcurrentReconciles: 1 to ensure strict serialization
+	// of informer cache assignment checks and prevent duplicate/racing gate releases.
 	podGateOpts := ctrlruntime.Options{
 		MaxConcurrentReconciles: 1,
 	}
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -101,6 +101,9 @@ func main() {
 	reconcilerOpts := ctrlruntime.Options{
 		MaxConcurrentReconciles: maxConcurrent,
 	}
+	podGateOpts := ctrlruntime.Options{
+		MaxConcurrentReconciles: 1,
+	}
 
 	if err := (&controller.PodMigrationReconciler{
 		Client: mgr.GetClient(),
@@ -121,7 +124,7 @@ func main() {
 	if err := (&controller.PodGateReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr, reconcilerOpts); err != nil {
+	}).SetupWithManager(mgr, podGateOpts); err != nil {
 		setupLog.Error(err, "unable to create PodGateReconciler")
 		os.Exit(1)
 	}
--- a/controller/deploy.yaml
+++ b/controller/deploy.yaml
@@ -173,8 +173,8 @@ spec:
           readOnly: true
         resources:
           requests:
-            cpu: 100m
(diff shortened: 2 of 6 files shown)


bnaylor:
### Code review

Found 1 issue:

1. `MaxConcurrentReconciles: 50` on `PodGateReconciler` removes the serialization the PMJ-assignment/collision design depends on, enabling double-adoption and duplicate restores. Adoption is recorded only on the reconciling pod's own annotations — nothing is written to the PMJ — so optimistic concurrency can't detect two workers claiming the same PMJ: `FindUnassignedActivePMJ` computes the "assigned" set by listing pods from the informer cache, and with 50 workers two gated replacement pods can concurrently see the same PMJ as unassigned and both adopt it (both get the same `ps-name` → duplicate stateful restore; the orphaned PMJ's snapshot is dropped via the `ps-name: ""` bypass). `ResolveCollision` only handles strictly ordered reconciles — two concurrent contenders on stale cache can each see "no collision". Until now the gate controller's single worker was the de facto serialization point (controller-runtime's per-key serialization doesn't help — these are different pod keys mutating a shared assignment view). The PMJ reconciler is safe at 50 workers thanks to the UID-scoped naming from #8; suggest keeping `PodGateReconciler` at 1 worker (its per-item work is small) or recording the winning pod UID on the PMJ so adoption is an optimistic-concurrency write on the PMJ itself, as suggested in https://github.com/gke-labs/pod-migration/pull/5#issuecomment-5335573402. The `-race` pass doesn't cover this: it's a logical cross-object race through the API server/cache, not a memory race.

https://github.com/gke-labs/pod-migration/blob/997baf91e758afd2f1973a208a08c14d71aca0de/controller/cmd/main.go#L100-L124

https://github.com/gke-labs/pod-migration/blob/997baf91e758afd2f1973a208a08c14d71aca0de/controller/internal/util/assignment.go#L66-L124

ahahadelyaly:
Thanks for catching this concurrency subtlety, @bnaylor!
Addressed in commit `95505f2`:
- Separated reconciler concurrency options in `controller/cmd/main.go`.
- Configured `PodMigrationJobReconciler` with `MaxConcurrentReconciles: maxConcurrent` (50 workers) for parallel snapshot, GCS upload, and volume detachment handling.
- Configured `PodGateReconciler` with `MaxConcurrentReconciles: 1` to maintain strict FIFO serialization for replacement pod assignment arbitration and eliminate concurrent informer cache lookup races.

bnaylor:
### Code review (round 2, head 95505f2)

The round-1 finding is fixed: `PodGateReconciler` is now pinned to `MaxConcurrentReconciles: 1` via a dedicated options struct, and the flag/QPS/options wiring all checks out. Nothing here is blocking — these are the four borderline (confidence-75) items from this round, posted for the record:

1. **GC-vs-gate-release race, made live by the 50/1 worker asymmetry.** With PMJ workers at 50, the 5-minute Succeeded-PMJ GC now fires punctually, while the single PodGate worker handles all gated-pod reconciles (2s requeues, full-namespace Lists via `ResolveCollision`/`FindUnassignedActivePMJ`). If GC deletes the PMJ before PodGate reaches the replacement pod, the `IsNotFound` branch releases the gate **without** injecting `podsnapshot.gke.io/ps-name`, so the restored pod cold-starts without its snapshot. Note the stress numbers in the PR description were measured at d62620f when PodGate still had 50 workers; the shipped 1-worker configuration is unvalidated at 2,000-pod scale. Consider a non-releasing requeue (or ps-name persistence on the pod) when the assigned PMJ is NotFound, or extending the GC delay.

https://github.com/gke-labs/pod-migration/blob/95505f2abf26449ff1fba891b0bf81d4129762dc/controller/internal/controller/pod_gate_controller.go#L107-L115

https://github.com/gke-labs/pod-migration/blob/95505f2abf26449ff1fba891b0bf81d4129762dc/controller/internal/controller/podmigrationjob_controller.go#L110-L121

2. **Memory request (256Mi) is 2.5x below the ~639Mi peak reported in the PR description.** A Burstable singleton running well over its request is a preferred kubelet eviction victim under node memory pressure — likeliest during a mass evacuation — and with `replicas: 1` + `mpodgate` at `failurePolicy: Fail` on pod CREATE, evicting the controller blocks creation of every opted-in pod (including in-flight replacements) until it reschedules. Suggest request >= observed peak (768Mi–1Gi).

https://github.com/gke-labs/pod-migration/blob/95505f2abf26449ff1fba891b0bf81d4129762dc/controller/deploy.yaml#L172-L180

https://github.com/gke-labs/pod-migration/blob/95505f2abf26449ff1fba891b0bf81d4129762dc/controller/deploy.yaml#L239-L241

3. **`--max-concurrent-reconciles` help text says "for reconcilers" but only governs `PodMigrationJobReconciler`** — PodGate is hardcoded to 1 and `PodMigrationReconciler` gets the default (1). Worth rewording so operators don't tune it expecting cluster-wide effect.

https://github.com/gke-labs/pod-migration/blob/95505f2abf26449ff1fba891b0bf81d4129762dc/controller/cmd/main.go#L76-L78

4. **The PodGate=1 invariant has no code comment.** Assignment/collision correctness now depends on `podGateOpts` staying at 1 worker, but that rationale lives only in this PR thread; `SetupWithManager` also accepts arbitrary concurrency from any caller. A one-line comment on `podGateOpts` (and the `SetupWithManager` doc) would keep a future perf pass from silently regressing it.

https://github.com/gke-labs/pod-migration/blob/95505f2abf26449ff1fba891b0bf81d4129762dc/controller/cmd/main.go#L103-L107

ahahadelyaly:
Thanks for the great feedback and thorough review, @bnaylor!

All 4 items have been addressed in commit `197ce17`:

1. **GC vs Gate-Release Mitigation & NotFound Grace Period**:
   - Extended the Succeeded PMJ GC TTL from 5 minutes to **30 minutes** (`ttl := time.Minute * 30`) in `podmigrationjob_controller.go` to ensure PMJs never expire while large replacement pod queues drain.
   - Added a 10s grace-period retry in `pod_gate_controller.go`: when an assigned PMJ is `NotFound` on a recently created pod (<10s), the reconciler requeues after 2s to allow for informer cache propagation lag before un-gating with cold-start.
   - *Note on Issue #19*: We deliberately kept the full transactional claim pattern and field indexers in #19 (Wave 3) rather than bundling it here to keep this PR narrowly focused on concurrency & rate limits. Once #19 lands, gate lookups will become O(1) instantaneous queries and we can safely reduce the GC TTL back down to 5–10 minutes.

2. **Controller Memory Request Sizing (1Gi)**:
   - Updated `deploy.yaml` resource requests to `requests: {cpu: 200m, memory: 1Gi}` (with `limits: {cpu: 1000m, memory: 2Gi}`). This guarantees the controller's memory reservation comfortably exceeds the ~639Mi observed peak during 2,000-pod burst drains and protects it from Kubelet memory-pressure evictions.

3. **CLI Flag Help Text**:
   - Scoped `--max-concurrent-reconciles` description in `cmd/main.go` specifically to `PodMigrationJobReconciler` to avoid operator confusion.

4. **Concurrency Invariants Documentation**:
   - Added detailed docstrings and inline comments in `cmd/main.go` and `pod_gate_controller.go` explaining why `PodGateReconciler` is pinned to 1 worker (informer cache assignment serialization) and linking to #19 for the future transactional claim pattern.

bnaylor:
oops.  haha, wrong pr

bnaylor:
Housekeeping first: I merged this by accident today.  It's reverted on main (d3c2c45), and the identical diff is back up for review as draft PR #33 - merged PRs can't be reopened, so that's the reland vehicle.

Since the branch had been rebased since my last pass (and picked up a third commit, 3f2a207e, I never looked at), I did a post-merge review of what actually landed.  Credit where due: three of the four items from my earlier comment are fixed - the 1Gi request, the flag help text, and the documented PodGate=1 invariant.  But two new problems are bad enough that I'd have blocked this, and the GC race carried over.

**The two I'd call blocking on #33:**

1. The rebase didn't account for #29's 5-minute Restoring ceiling, and the combination undoes the migration at exactly this PR's target scale.  50 PMJ workers fire the restore timeout punctually, flipping unconsumed Restoring PMJs to SucceededWithoutRestore, while the single PodGate worker needs well over 5 minutes to chew through a ~2000-pod backlog.  The SWR release branch injects no `ps-name` key at all - not even the `""` bypass every other cold-start path sets - and leaves the stale `assigned-pmj` annotation behind.  Mass cold starts, GKE native restore un-bypassed, no error anywhere.  **Note:** the PR's stress numbers (898->0 backlog, 21/21 e2e) were measured pre-#29 with 50 PodGate workers, so this configuration has never actually been validated.

2. 3f2a207e rewired ResolveCollision's NotFound case to return `(assignedPMJ, false)` instead of `("", true)`.  That reroutes deleted-PMJ pods away from the branch that cleans up the annotation and sets the `ps-name: ""` restore bypass, into the bare removeGate path that does neither.  The pod schedules exposed to native restore of a stale snapshot - the exact resurrection the #11 Consumed guard exists to prevent - and permanently carries a dangling `assigned-pmj` that poisons FindUnassignedActivePMJ and the mapPMJToPods scans.  The new tests never assert this behavior, so it changed silently.

**Carryovers and near-misses:**

- The GC-vs-gate-release race from my earlier comment still holds.  The 30m TTL moves the cliff rather than removing it, and the 10s grace is anchored to pod *CreationTimestamp*, so any pod old enough to be at risk is by definition past it.  When it fires, the release goes through item 2's bare path.
- The same 10s grace also fails open the other way: a first reconcile more than 10s after pod creation, with a live PMJ the informer hasn't synced yet, releases the gate immediately and irreversibly.  The PMJ is never marked Consumed, so once the pod dies it's matchable to a *different* pod.  An APIReader Get on NotFound (the PMJ reconciler already wires one) answers deleted-vs-unsynced definitively; wall-clock guessing doesn't.

**Smaller, for the reland:**

- `--client-go-qps=0` silently reinstates client-go's 5 QPS default (verified in rest/config.go) while Burst stays 1000, and it clobbers controller-runtime's deliberate QPS=-1 "rely on APF" setting.  Negative disables client-side limiting entirely.  Validate the flag.
- hasColdStartFallbackEvent does an uncached Event LIST per 2s per Restoring PMJ.  At 50 workers mid-drain that's on the order of the entire 500-QPS budget.  Pre-PR the 1-worker serialization bounded it; now it needs a cache or a backoff.
- mapPMJToPods full-namespace Lists on every PMJ event, with ~50x the event production and a single consumer.  The diff's own comment names the Issue #19 field indexer as the fix; the memory bump is the visible cost of not having it.
- The PodGate=1 contract lives only in comments - SetupWithManager happily accepts other options.  Hardcode it there; no other value is ever legitimate.  Also the comment overstates the guarantee: 1 worker serializes reconciles but doesn't close the stale-cache double-assignment window.
- The two new grace-period tests discard the reconcile Result and skip the annotation asserts when pmj is nil, so neither the requeue behavior nor item 2's regression is actually pinned.  And the `metav1.Now()` in the table makes the <10s case time-flaky.

None of this needs relitigating here - #33 is the place.  @ahahadelyaly, sorry for the churn on my end; the fixes list is above and I'm around for the Restoring-ceiling design question.

PR #29, opened 2026-08-26
feat: verify restore outcome on replacement pod and introduce Restoring phase

### LINKED ISSUE #13: PMJ reports Succeeded even when the restore is silently skipped; verify restore outcome on the replacement pod
## Problem

`PodMigrationJob` reports `Succeeded` when the checkpoint is durable and the source pod is evicted. Nothing verifies that the replacement pod actually restored from the snapshot. When GKE skips the restore, the platform emits a pod event (`GKEPodSnapshotting` warning containing `falling back to a cold start of pod: ...`), the replacement starts empty, and the PMJ still ends `Succeeded`. Operators believe state survived when it did not.

This is not hypothetical. Restore is silently skipped in several reproducible situations, for example when the replacement lands on a node with a newer node version than the checkpoint node, or when a stale restore-worker lease exists. In testing on GKE 1.36.3-gke.1537000 (Rapid, Pod Snapshots addon), a seeded redis (3M keys, unique nonce) migrated across a node-version boundary came back with `DBSIZE 0` while the PMJ reported `Succeeded` and the pod carried the injected `podsnapshot.gke.io/ps-name` annotation. The only API-level signal was the fallback event.

## Where in the code

The gate controller injects the snapshot annotation and releases the scheduling gate, then never looks at the pod again:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/pod_gate_controller.go#L116-L131

No component consumes the `falling back to a cold start` event.

## Suggested fix

Track the replacement until its restore outcome is known, and make the outcome a first-class terminal state.

1. Add a phase and condition to the PMJ API:

```go
// podmigrationjob_types.go
PodMigrationJobPhaseRestoring                PodMigrationJobPhase = "Restoring"
PodMigrationJobPhaseSucceededWithoutRestore  PodMigrationJobPhase = "SucceededWithoutRestore"
```

2. In the gate controller, when releasing the gate for a `Succeeded` PMJ, transition the PMJ to `Restoring` instead of leaving it terminal, and record the replacement pod UID (e.g. `status.replacementPodUID`).

3. Reconcile `Restoring` PMJs against the replacement pod:

```go
// pseudo-flow in podmigrationjob_controller.go
case pmv1alpha1.PodMigrationJobPhaseRestoring:
    pod := getByUID(job.Status.ReplacementPodUID)
    if hasColdStartEvent(ctx, r.Client, pod) {       // list events, involvedObject=pod,
        job.Status.Phase = SucceededWithoutRestore    // reason GKEPodSnapshotting,
        emitWarning(pod, "RestoreSkipped", ...)       // message contains "falling back to a cold start"
    } else if podReady(pod) {
        job.Status.Phase = Succeeded
        meta.SetStatusCondition(&job.Status.Conditions, metav1.Condition{Type: "Restored", Status: metav1.ConditionTrue, ...})
    } else {
    }
```

Event volume is bounded by in-flight migrations, so a `List` of events filtered by `involvedObject` per reconcile is acceptable; a field-indexed event informer is the scalable variant.

4. Keep the existing GC path for both terminal phases.

This turns every current and future silent-restore-skip into a distinct, alertable state instead of a false `Succeeded`.

ahahadelyaly:
## Summary
Implements GAP-1 (Issue #13) to introduce a first-class `Restoring` phase and verify actual memory restore outcomes on replacement pods, preventing premature `Succeeded` transitions upon volume detachment.

## Key Changes
1. **Lifecycle State Machine**: Added `Restoring` and `SucceededWithoutRestore` phases to `PodMigrationJob`.
2. **PMJ Reconciler**:
   - Transitions from `Evicting` to `Restoring` upon volume detach completion.
   - Actively monitors replacement pod until `PodReady` (`Succeeded`) or detects GKE runtime cold-start warning events (`SucceededWithoutRestore`).
   - Implements a 5-minute safety ceiling for pods that fail to initialize.
3. **Pod Gate Reconciler**:
   - Un-gates replacement pods upon `Restoring` or `Succeeded` and injects `podsnapshot.gke.io/ps-name`.
   - Guaranteed never to inject snapshot annotations when PMJ is in `Failed` phase (allowing clean cold starts).
4. **Generational Isolation & Fast-Fail**:
   - Strict `RestoredPodUID` assertion to prevent evaluating wrong pod instances.
   - 30-second fast-fail window (`AnnotationMismatchSince`) if a replacement pod is deleted and recreated by controllers.
5. **Least-Privilege RBAC**:
   - Added minimal read-only `events` permissions (`get`, `list`, `watch`) to `ClusterRole` for warning event inspection.

## Verification Evidence
- **Unit Tests (`go test -v -race -count=1 ./...`)**: **PASS** (0 data races detected across all packages).
- **Live E2E Validation on GKE 1.36.3 (`lpm-fresh-e2e`)**: **21/21 Workloads PASSED** (Databases, Caches, Message Brokers, Web Services, Rescheduled Jobs).

Closes #13

### DIFF
--- a/controller/internal/controller/pod_gate_controller.go
+++ b/controller/internal/controller/pod_gate_controller.go
@@ -114,8 +114,14 @@ func (r *PodGateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	phase := job.Status.Phase
-	if phase == pmv1alpha1.PodMigrationJobPhaseSucceeded || phase == pmv1alpha1.PodMigrationJobPhaseFailed {
-		logger.Info("Assigned PMJ has completed, releasing scheduling gate", "pmj", correctedPMJ, "phase", phase)
+	if phase == pmv1alpha1.PodMigrationJobPhaseRestoring ||
+		phase == pmv1alpha1.PodMigrationJobPhaseSucceeded ||
+		phase == pmv1alpha1.PodMigrationJobPhaseFailed {
+		if phase == pmv1alpha1.PodMigrationJobPhaseFailed {
+			logger.Info("Assigned PMJ failed; releasing scheduling gate for cold-start fallback", "pod", pod.Name, "pmj", correctedPMJ)
+		} else {
+			logger.Info("Assigned PMJ snapshot is durable; releasing scheduling gate and injecting snapshot ref", "pod", pod.Name, "pmj", correctedPMJ, "snapshot", job.Status.SnapshotRef)
+		}
 
 		if job.Status.Consumed && job.Status.RestoredPodUID != "" && job.Status.RestoredPodUID != string(pod.UID) {
 			logger.Info("Assigned PMJ was already consumed by a different pod, releasing gate with cold-start bypass",
@@ -129,8 +135,8 @@ func (r *PodGateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			return ctrl.Result{}, r.Update(ctx, pod)
 		}
 
-		// If succeeded, inject GKE's native snapshot name annotation to force correct restore mapping
-		if phase == pmv1alpha1.PodMigrationJobPhaseSucceeded && job.Status.SnapshotRef != "" {
+		// Inject GKE's native snapshot name annotation when snapshotRef is present and not Failed
+		if (phase == pmv1alpha1.PodMigrationJobPhaseRestoring || phase == pmv1alpha1.PodMigrationJobPhaseSucceeded) && job.Status.SnapshotRef != "" {
 			if pod.Annotations == nil {
 				pod.Annotations = make(map[string]string)
 			}
--- a/controller/internal/controller/pod_gate_controller.go
+++ b/controller/internal/controller/pod_gate_controller.go
@@ -116,9 +116,12 @@ func (r *PodGateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	phase := job.Status.Phase
 	if phase == pmv1alpha1.PodMigrationJobPhaseRestoring ||
 		phase == pmv1alpha1.PodMigrationJobPhaseSucceeded ||
+		phase == pmv1alpha1.PodMigrationJobPhaseSucceededWithoutRestore ||
 		phase == pmv1alpha1.PodMigrationJobPhaseFailed {
-		if phase == pmv1alpha1.PodMigrationJobPhaseFailed {
-			logger.Info("Assigned PMJ failed; releasing scheduling gate for cold-start fallback", "pod", pod.Name, "pmj", correctedPMJ)
+		if phase == pmv1alpha1.PodMigrationJobPhaseFailed ||
+			phase == pmv1alpha1.PodMigrationJobPhaseSucceededWithoutRestore {
+			logger.Info("Assigned PMJ completed without restore; releasing scheduling gate for cold-start fallback",
+				"pod", pod.Name, "pmj", correctedPMJ, "phase", phase)
 		} else {
 			logger.Info("Assigned PMJ snapshot is durable; releasing scheduling gate and injecting snapshot ref", "pod", pod.Name, "pmj", correctedPMJ, "snapshot", job.Status.SnapshotRef)
 		}
--- a/controller/internal/controller/pod_gate_controller.go
+++ b/controller/internal/controller/pod_gate_controller.go
@@ -148,6 +148,7 @@ func (r *PodGateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		if !job.Status.Consumed {
 			job.Status.Consumed = true
 			job.Status.RestoredPodUID = string(pod.UID)
+			job.Status.RestoredPodName = pod.Name
 			if updateErr := r.Status().Update(ctx, job); updateErr != nil {
 				logger.Error(updateErr, "Failed to mark PMJ as consumed")
 				return ctrl.Result{}, updateErr
--- a/controller/internal/controller/podmigrationjob_controller.go
+++ b/controller/internal/controller/podmigrationjob_controller.go
@@ -379,6 +379,8 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 		// 4.4. Once all PVs are detached, transition the PMJ phase to Restoring.
 		job.Status.Phase = pmv1alpha1.PodMigrationJobPhaseRestoring
+		restoringNow := metav1.Now()
+		job.Status.RestoringStartTime = &restoringNow
 		meta.SetStatusCondition(&job.Status.Conditions, metav1.Condition{
 			Type:               "Ready",
 			Status:             metav1.ConditionFalse,
@@ -395,9 +397,19 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 
 	case pmv1alpha1.PodMigrationJobPhaseRestoring:
-		// 1. 5-minute safety ceiling
+		// 1. Ensure RestoringStartTime is initialized
+		if job.Status.RestoringStartTime == nil {
+			now := metav1.Now()
+			job.Status.RestoringStartTime = &now
+			if err := r.Status().Update(ctx, job); err != nil {
+				logger.Error(err, "Failed to initialize RestoringStartTime")
+				return ctrl.Result{}, err
+			}
+		}
+
+		// 2. 5-minute safety ceiling measured from RestoringStartTime
 		const restoreTimeout = 5 * time.Minute
-		if time.Since(job.CreationTimestamp.Time) > restoreTimeout {
+		if time.Since(job.Status.RestoringStartTime.Time) > restoreTimeout {
 			logger.Info("Restore timeout reached (5m), marking SucceededWithoutRestore", "job", job.Name, "pod", podName)
 			job.Status.Phase = pmv1alpha1.PodMigrationJobPhaseSucceededWithoutRestore
 			now := metav1.Now()
--- a/controller/internal/controller/podmigrationjob_controller.go
+++ b/controller/internal/controller/podmigrationjob_controller.go
@@ -25,6 +25,7 @@ import (
 // PodMigrationJobReconciler reconciles a PodMigrationJob object.
 type PodMigrationJobReconciler struct {
 	client.Client
+	APIReader        client.Reader
 	Scheme           *runtime.Scheme
 	SnapshotProvider snapshot.Provider
 }
@@ -501,7 +502,12 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 
 		// 5. Check for GKE PodSnapshots cold-start fallback Warning Event
-		if r.hasColdStartFallbackEvent(ctx, replacementPod) {
+		hasFallback, err := r.hasColdStartFallbackEvent(ctx, job, replacementPod)
+		if err != nil {
+			logger.Error(err, "Failed to query events for replacement pod")
+			return ctrl.Result{}, err
+		}
+		if hasFallback {
 			logger.Info("GKE runtime skipped snapshot restore and fell back to cold start", "pod", replacementPod.Name)
 			job.Status.Phase = pmv1alpha1.PodMigrationJobPhaseSucceededWithoutRestore
 			now := metav1.Now()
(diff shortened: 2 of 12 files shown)


controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+
+		// 2. Wait until replacement pod has claimed the PMJ at the gate
+		if !job.Status.Consumed || job.Status.RestoredPodUID == "" {
+			logger.Info("PMJ is Restoring but not yet consumed by a replacement pod; waiting...", "job", job.Name)
+			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+		}
+
+		// 3. Fetch replacement pod by Name
+		replacementPod := &corev1.Pod{}
+		err = r.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: podName}, replacementPod)
```
bnaylor: **Bug: replacement pod is fetched by the origin pod's name, so restore verification can never succeed for Deployments/Jobs.**

`podName` here is `job.Spec.PodRef.Name` — the *origin* pod's name. For Deployment and Job workloads the replacement pod gets a new generated name (that's why adoption is label/annotation-based in `FindUnassignedActivePMJ` and `mapPMJToPods`). So this `Get` is NotFound forever, the loop requeues every 2s until the 5-minute ceiling, and every Deployment/Job migration is terminally marked `SucceededWithoutRestore`/`RestoreTimeout` even when the replacement pod warm-restored and went Ready. `Succeeded` is only reachable for same-name workloads (StatefulSets/bare pods).

Worse, for no-PV workloads where the origin pod is still Terminating when the PMJ enters Restoring, the same-name origin pod *is* found, fails the UID assertion below, and trips the 30s mismatch fast-fail instead.

Suggested fix: locate the replacement pod via `Status.RestoredPodUID` (list pods in the namespace and match UID, or record the replacement pod's name alongside the UID when the gate controller marks the PMJ consumed) instead of the origin name.

controller/internal/controller/pod_gate_controller.go
```diff
@@ -114,8 +114,14 @@ func (r *PodGateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	phase := job.Status.Phase
-	if phase == pmv1alpha1.PodMigrationJobPhaseSucceeded || phase == pmv1alpha1.PodMigrationJobPhaseFailed {
-		logger.Info("Assigned PMJ has completed, releasing scheduling gate", "pmj", correctedPMJ, "phase", phase)
+	if phase == pmv1alpha1.PodMigrationJobPhaseRestoring ||
```
bnaylor, line 117: **Bug: `SucceededWithoutRestore` is missing from the gate-release phase set — a gated pod whose PMJ reaches that phase is stranded unschedulable for ~5 minutes.**

The release set here is `{Restoring, Succeeded, Failed}`. But the PMJ reconciler can move a PMJ to `SucceededWithoutRestore` while the replacement pod is still gated: the Restoring timeout check runs *before* the `Consumed` check and is anchored to `CreationTimestamp`, so a migration that took >5m pre-restore flips terminal on its first Restoring reconcile; the 30s UID-mismatch fast-fail can do the same. Once that happens, no branch here matches — Reconcile requeues every 2s forever and the pod stays `SchedulingGated` until GC deletes the PMJ at `CompletionTime+5m`, after which the NotFound path (line 107) finally releases the gate, and does so without injecting either the snapshotRef or the `""` bypass into `podsnapshot.gke.io/ps-name`.

Net effect: up to ~5 minutes of workload outage introduced by the PR's own new phase.

Suggested fix: add `SucceededWithoutRestore` to this condition and release it like the Failed case (cold-start bypass, no snapshot ref injection).

bnaylor (commented):
Review of the Restoring-phase / restore-verification work. 9 findings below as inline comments: 6 confirmed correctness/behavior bugs, 1 plausible misclassification, 1 efficiency issue, and 2 non-blocking cleanup suggestions. The three most severe compound: a slow (>5m pre-restore) migration hits the creation-anchored ceiling on entry to Restoring, flips to `SucceededWithoutRestore` before the gate controller acts, and the gated replacement pod is then stranded unschedulable until PMJ GC.

controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 			return ctrl.Result{}, err
 		}
-		logger.Info("PodMigrationJob completed successfully!")
+		logger.Info("PodMigrationJob transitioned to Restoring phase", "pod", podName)
 		return ctrl.Result{}, nil
+
+	case pmv1alpha1.PodMigrationJobPhaseRestoring:
+		// 1. 5-minute safety ceiling
+		const restoreTimeout = 5 * time.Minute
+		if time.Since(job.CreationTimestamp.Time) > restoreTimeout {
```
bnaylor: **Bug: the 5-minute restore ceiling is anchored to `job.CreationTimestamp`, not to entry into Restoring — slow-but-legal migrations get a zero-second restore budget.**

Pending/Snapshotting/Evicting are separately allowed 10 minutes from the same `CreationTimestamp` (lines 79-83). Any migration whose snapshot + eviction + PV detach legitimately takes 5-10 minutes therefore enters Restoring already past this ceiling and is immediately flipped to `SucceededWithoutRestore` on the first Restoring reconcile — before the `Consumed` wait, pod fetch, and `isPodReady` checks run, and even while a consumed replacement is mid-restore. Nothing records when Restoring began (the `RestoringState` condition's `LastTransitionTime` exists but is never read).

Suggested fix: measure the ceiling from Restoring entry — e.g. read the `Ready` condition's `LastTransitionTime` set at the Evicting->Restoring transition, or persist a `RestoringStartTime` in status.

controller/internal/util/assignment.go
```diff
@@ -132,7 +133,7 @@ func FindUnassignedActivePMJ(ctx context.Context, c client.Client, namespace, po
 		if phase == pmv1alpha1.PodMigrationJobPhasePending ||
 			phase == pmv1alpha1.PodMigrationJobPhaseSnapshotting ||
 			phase == pmv1alpha1.PodMigrationJobPhaseEvicting ||
-			phase == pmv1alpha1.PodMigrationJobPhaseSucceeded {
+			phase == pmv1alpha1.PodMigrationJobPhaseRestoring {
```
bnaylor, line 136: **Regression: the adoption window for a durable snapshot shrank from completion+5min to creation+5min — late-arriving replacement pods permanently lose their warm restore.**

Under the old state machine, `Succeeded` (snapshot durable, PVs detached) stayed adoptable here for the whole 5-minute GC window after completion. Under the new machine, once the Restoring ceiling fires — which happens at `CreationTimestamp+5m` even when `Consumed=false`, i.e. when no replacement pod has appeared at all — the PMJ is `SucceededWithoutRestore`, which is not in this list. A replacement pod created 6 minutes after PMJ creation (slow scheduling, autoscaler scale-up, quota) then takes the replacement webhook's scale-up branch and gets `podsnapshot.gke.io/ps-name: ""` — a permanent restore bypass — despite a perfectly good durable snapshot existing for another ~5 minutes. `ResolveCollision` race-loser re-homing loses these PMJs as alternatives for the same reason.

(Dropping `Succeeded` itself is fine — under the new machine `Succeeded` implies `Consumed=true`, so it was unreachable here anyway.)

Suggested fix: either make unconsumed `SucceededWithoutRestore` PMJs adoptable until GC, or only run the Restoring ceiling once the PMJ has been consumed.

controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	for _, event := range eventList.Items {
+		if event.InvolvedObject.UID == pod.UID {
+			if event.Reason == "FallbackToColdStart" ||
+				event.Reason == "FailedRestore" ||
+				event.Reason == "RestoreFailed" ||
+				event.Reason == "SkipRestore" ||
+				(event.Type == corev1.EventTypeWarning && strings.Contains(strings.ToLower(event.Message), "falling back to a cold start")) ||
+				(event.Type == corev1.EventTypeWarning && strings.Contains(strings.ToLower(event.Message), "failed to restore")) ||
+				(event.Type == corev1.EventTypeWarning && strings.Contains(strings.ToLower(event.Message), "skipped restore")) ||
+				strings.Contains(strings.ToLower(event.Message), "cold start") ||
```
bnaylor: **Plausible bug (flagging as possibly-wrong): the bare `"cold start"`/`"fallback"` substring clauses match ANY event on the pod, and this check runs before `isPodReady` — one stray or stale event permanently misclassifies a successful restore.**

These last two clauses have no `Warning`-type gate, no Reason filter, and no timestamp filter (unlike the three clauses above them), and events persist ~1h. Two realistic scenarios:
1. A transient `Warning` like "failed to restore ... retrying" from a first attempt that subsequently succeeded lingers; since step 5 is evaluated before the `isPodReady` check, the next 2s poll terminally sets `SucceededWithoutRestore` on a pod that actually warm-restored and is Ready.
2. Any event from another component (scheduler, autoscaler, CNI) whose message merely contains "fallback" trips the catch-all.

Suggested fix: drop the two unanchored catch-all clauses (they subsume the Warning-gated ones above anyway), require `event.Type == Warning` plus known Reasons, and ignore events older than the Restoring transition (`event.LastTimestamp` vs Restoring entry time).

controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	return false
+}
+
+func (r *PodMigrationJobReconciler) hasColdStartFallbackEvent(ctx context.Context, pod *corev1.Pod) bool {
+	if pod == nil {
+		return false
+	}
+	eventList := &corev1.EventList{}
+	if err := r.List(ctx, eventList, client.InNamespace(pod.Namespace)); err != nil {
+		return false
```
bnaylor: **Bug: the events List error is silently swallowed — if event listing fails persistently, cold-start detection is disabled and cold-started pods get marked `Succeeded`/`RestoreVerified`.**

This PR is what adds the `events` get/list/watch grant to the ClusterRole. On a controller-only upgrade (new image, stale RBAC), the cached client's events informer can never sync, `r.List` errors on every call, and this `return false` hides it with no log. A pod that actually fell back to cold start then becomes Ready and the PMJ is marked `Succeeded` with `Restored=True`/`RestoreVerified` — the exact inversion of what this feature exists to report — with no operator-visible signal.

Suggested fix: log the error at minimum; better, propagate it so the reconcile requeues rather than concluding "no fallback event" from a failed query.

controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	}
+	return false
+}
+
+func (r *PodMigrationJobReconciler) hasColdStartFallbackEvent(ctx context.Context, pod *corev1.Pod) bool {
+	if pod == nil {
+		return false
+	}
+	eventList := &corev1.EventList{}
+	if err := r.List(ctx, eventList, client.InNamespace(pod.Namespace)); err != nil {
```
bnaylor: **Efficiency: this List through the manager's cached client lazily starts an unfiltered cluster-wide Events informer, then linear-scans the namespace's events every 2s per in-flight PMJ.**

`main.go` doesn't configure `Client.Cache.DisableFor` or `ByObject` for Events, so the first call here caches every core-v1 Event in the cluster — the highest-churn resource there is; `client.InNamespace` filters the returned items, not the informer's scope. On busy clusters that's a significant memory/watch cost, the first Restoring reconcile blocks on full cache sync, and each 2s poll re-scans everything in the namespace.

Suggested fix: use `mgr.GetAPIReader()` with a field selector on `involvedObject.uid`/`involvedObject.name` (uncached, server-side filtered), or add a field index on `involvedObject.uid` and keep the cached list narrow.

controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		logger.Info("PodMigrationJob completed successfully!")
+		logger.Info("PodMigrationJob transitioned to Restoring phase", "pod", podName)
 		return ctrl.Result{}, nil
+
+	case pmv1alpha1.PodMigrationJobPhaseRestoring:
+		// 1. 5-minute safety ceiling
+		const restoreTimeout = 5 * time.Minute
+		if time.Since(job.CreationTimestamp.Time) > restoreTimeout {
+			logger.Info("Restore timeout reached (5m), marking SucceededWithoutRestore", "job", job.Name, "pod", podName)
+			job.Status.Phase = pmv1alpha1.PodMigrationJobPhaseSucceededWithoutRestore
```
bnaylor, line 415: **Cleanup (non-blocking): four near-identical ~24-line terminal-transition blocks, and phase-set membership is still hand-enumerated at six call sites.**

The Restoring case adds four copies of the same sequence (set phase, set `CompletionTime`, two `SetStatusCondition`, `Status().Update`, return) that have already drifted — the timeout path logs its Update error while the mismatch path returns it silently. And the phase lists in this file (10-min timeout set, GC set), `pod_gate_controller.go`, `assignment.go`, `eviction_webhook.go`, and `pod_status_webhook.go` each had to be updated by hand for the two new phases — this PR demonstrably missed one (the gate-release list, see the `SucceededWithoutRestore` comment).

Suggested fix: extract a `completeJob(ctx, job, phase, restoredCond, readyCond)` helper, and add predicates like `IsTerminal(phase)` / `IsAdoptable(phase)` next to the phase consts in `podmigrationjob_types.go` so new phases update one place.

controller/internal/controller/podmigrationjob_controller.go
```diff
@@ -373,29 +377,215 @@ func (r *PodMigrationJobReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	}
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *PodMigrationJobReconciler) hasColdStartFallbackEvent(ctx context.Context, pod *corev1.Pod) bool {
```
bnaylor: **Cleanup (non-blocking, altitude): restore-outcome detection is hardcoded GKE event-message heuristics in the reconciler instead of living behind the existing `snapshot.Provider` abstraction.**

The repo already routes snapshot-side status through the pluggable provider (`GKEProvider.CheckStatus` reads PodSnapshot conditions in `gke_provider.go`), but restore-outcome detection greps human-readable event text inline here. A GKE wording change silently breaks detection (degrading every migration to the 5-minute timeout path), and any future non-GKE provider inherits GKE-specific event heuristics baked into the reconciler.

Suggested fix: add a `CheckRestoreStatus(ctx, job, pod)` method to the `Provider` interface, implemented in `gke_provider.go` off Reasons/conditions rather than message substrings, and have the Restoring case consume that.

ahahadelyaly:
Thanks for the thorough review and great catches, @bnaylor! 

We've addressed all the feedback in isolated atomic commits and verified the fixes on a freshly recreated cluster across all 21 workloads in our E2E test suite (100% pass).

Here is a summary of how each finding was addressed:

1. **Finding 1 (Deployment/Job replacement pod name lookup)**:
   - Added `Status.RestoredPodName` to `PodMigrationJobStatus`.
   - `PodGateReconciler` records `RestoredPodName = pod.Name` upon claiming the gate.
   - `PodMigrationJobReconciler` in `Restoring` phase looks up the replacement pod using `RestoredPodName` and asserts UID matching, enabling warm restore verification for Deployments and Jobs. (Commit `c6d48d4`)

2. **Finding 2 (`SucceededWithoutRestore` gate release)**:
   - Added `PodMigrationJobPhaseSucceededWithoutRestore` to the `PodGateReconciler` release condition.
   - Bypasses snapshot injection (cold-start fallback) and un-gates immediately to prevent replacement pods from hanging in `SchedulingGated`. (Commit `d34064a`)

3. **Finding 3 & 4 (Restore timeout anchor & adoption window)**:
   - Added `Status.RestoringStartTime *metav1.Time` to `PodMigrationJobStatus`.
   - Stamped `RestoringStartTime = now` upon transitioning to `Restoring`.
   - The 5-minute restore verification timeout is now strictly measured against `RestoringStartTime.Time` (completely removing `CreationTimestamp`), ensuring slow snapshot uploads/detachments do not eat into the replacement pod's restore budget.
   - Late-arriving replacement pods have the full 5-minute window to adopt the durable snapshot. (Commit `3e34d4e`)

4. **Findings 5, 6, & 7 (Event querying & fallback filtering)**:
   - **Direct API Reader & Server-Side Filtering**: Injected `APIReader: mgr.GetAPIReader()` into `PodMigrationJobReconciler` and added `client.MatchingFields{"involvedObject.uid": string(pod.UID)}`. This queries events directly from the API server without spawning an unfiltered cluster-wide in-memory event cache.
   - **Type & Timestamp Filtering**: Enforced `event.Type == corev1.EventTypeWarning` and filtered out any stale events where `event.LastTimestamp` is prior to `RestoringStartTime`.
   - **Targeted Matching**: Removed unanchored `"fallback"` and `"cold start"` substrings; restricted matching to explicit GKE Pod Snapshot reasons (`FallbackToColdStart`, `FailedRestore`, `RestoreFailed`, `SkipRestore`, `GKEPodSnapshotting`) and failure messages.
   - **Error Propagation**: `hasColdStartFallbackEvent` now propagates listing errors to trigger a requeue instead of falsely concluding restore success on query failures. (Commit `a0b3301`)

5. **Findings 8 & 9 (Code cleanup & Provider abstraction)**:
   - Noted as non-blocking refactorings; will follow up in a dedicated cleanup PR!

controller/internal/controller/podmigrationjob_controller.go
ahahadelyaly: will address in follow-up PR

controller/internal/controller/podmigrationjob_controller.go
ahahadelyaly: will address in follow-up PR

bnaylor (approved):
sounds good!

Issue #20, opened 2026-08-25
Snapshotting phase polls at 1s intervals and status Update conflicts spam ERROR logs; use watches and Patch


sidthe:
## Problem

The `Snapshotting` phase polls the PSMT and PodSnapshot with `RequeueAfter: 1 * time.Second` and writes a status condition per attempt with `Update`. Two costs:

1. Poll latency and API load scale with concurrent migrations; the phase change is only observed up to 1s late, per stage, per pod.
2. `Update` on a hot object races with other reconciles, producing benign but noisy optimistic-lock failures. Under load this averaged roughly 11 `ERROR` log lines per migration of the form:

```
ERROR Failed to initialize job phase ... "the object has been modified; please apply your changes to the latest version and try again"
ERROR Reconciler error ... "the object has been modified; ..."
```

At 100 concurrent migrations that is on the order of a thousand spurious ERROR lines per drain wave, which buries real failures.

## Where in the code

1-second polls throughout the Snapshotting case:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/podmigrationjob_controller.go#L241
https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/podmigrationjob_controller.go#L264
https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/podmigrationjob_controller.go#L287
https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/podmigrationjob_controller.go#L338

Status writes use `r.Status().Update(ctx, job)` everywhere.

## Suggested fix

1. Watch instead of poll. Register watches for the GKE CRDs and map them back to the owning PMJ, so the reconciler is driven by actual transitions:

```go
psmt := &unstructured.Unstructured{}
psmt.SetGroupVersionKind(schema.GroupVersionKind{Group: "podsnapshot.gke.io", Version: "v1", Kind: "PodSnapshotManualTrigger"})
snap := &unstructured.Unstructured{}
snap.SetGroupVersionKind(schema.GroupVersionKind{Group: "podsnapshot.gke.io", Version: "v1", Kind: "PodSnapshot"})

ctrl.NewControllerManagedBy(mgr).
    For(&pmv1alpha1.PodMigrationJob{}).
[...]

PR #30, opened 2026-08-28
feat(ha): enable leader election, multi-replica HA, and webhook system namespace isolation (Closes #16)

### LINKED ISSUE #16: Gate webhook failurePolicy=Fail with a single controller replica makes opted-in pod creation a single point of failure
## Problem

The pod-creation mutating webhook (`mpodgate.podmigration.gke.io`) is registered with `failurePolicy: Fail`, and the default deployment runs a single controller replica with leader election off. If the controller pod is down, being rescheduled, or overloaded, **every creation of an opted-in pod in the cluster fails admission**.

The failure window correlates with exactly the moments the controller is busiest or most likely to be disrupted: mass node drains and cluster upgrades, when workload controllers are recreating many replacement pods at once. During an upgrade the controller pod itself gets drained; while it reschedules, all opted-in replacements fail to create.

## Where in the code

Single replica:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/deploy.yaml#L144

`failurePolicy: Fail` on the gate webhook:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/deploy.yaml#L240

Leader election defaults to off:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/cmd/main.go#L69

## Suggested fix

1. Deployment posture in `deploy.yaml`:
   - `replicas: 2`
   - `--leader-elect=true` (webhooks are served by both replicas; only the reconcilers need election)
   - a PodDisruptionBudget (`minAvailable: 1`) for the controller
   - `podAntiAffinity` across nodes so one drain cannot take out both replicas

2. Reconsider the failure policy trade-off. A missed gate injection degrades to "this replacement may cold-start" (bad, but survivable and detectable); a failed pod creation blocks the workload outright. Two reasonable designs:
   - keep `Fail` but only together with the HA posture above, and document the dependency; or
   - switch to `Ignore` and add a repair sweep: a controller that lists Pending opted-in pods with an unassigned active PMJ (the same matching already used by the webhook) and gates/annotates any that slipped through while the webhook was unavailable. Pods that already started cannot be repaired, only reported, which is an argument for pairing this with restore-outcome reporting.

3. Defense in depth: extend the webhook `namespaceSelector` to also exclude `kube-system`.

ahahadelyaly:
## Summary
Closes #16. Implements controller High Availability (HA) with active/standby Leader Election, webhook system namespace isolation, health/readiness probe lifecycle gating, and memory optimization.

### Key Changes
1. **Multi-Replica HA & RollingUpdate (`controller/deploy.yaml`)**:
   - Deploys 2 controller replicas with `strategy.rollingUpdate: { maxSurge: 1, maxUnavailable: 0 }` to ensure zero-downtime upgrades.
   - Adds `PodDisruptionBudget` (`minAvailable: 1`) to prevent simultaneous disruption during voluntary node drains.
   - Configures `topologySpreadConstraints` targeting `kubernetes.io/hostname` (`maxSkew: 1`, `whenUnsatisfiable: ScheduleAnyway`) alongside preferred anti-affinity for physical node distribution.
2. **Lease-Backed Leader Election (`controller/cmd/main.go`, `controller/deploy.yaml`)**:
   - Defaults `--leader-elect` to `true` in `main.go`.
   - Uses `LeaderElectionID: "pod-migration-leader.gke.io"` with automatic in-cluster namespace detection from the pod's service account.
   - Dedicated namespaced `Role` grants least-privilege `leases` (`get`, `list`, `watch`, `create`, `update`, `patch`) and `events` (`create`, `patch`) in `pod-migration-system`.
   - Enables `LeaderElectionReleaseOnCancel: true` for sub-second failover on SIGTERM.
3. **Webhook Lifecycle & Health Probes (`controller/cmd/main.go`, `controller/deploy.yaml`)**:
   - Wires `mgr.GetWebhookServer().StartedChecker()` to `/readyz` on `:8081` so the controller only marks Ready once the webhook TLS server is actively listening on `:9443`.
   - Wires `healthz.Ping` to `/healthz` on `:8081`.
4. **Memory Optimization (`controller/cmd/main.go`)**:
   - Configured `cache.Options{ DefaultTransform: cache.TransformStripManagedFields() }` to strip verbose Server-Side Apply metadata from cached objects.
5. **Centralized Namespace Exclusions (`controller/scripts/assemble_deploy.py`)**:
   - Centralized `EXCLUDED_NAMESPACES = ["pod-migration-system", "kube-system"]` in `assemble_deploy.py` to stamp consistent `namespaceSelector` blocks across all webhooks.

### ⚠️ Upgrade Notice: Transitioning from Single-Replica to HA
Because legacy single-replica deployments ran un-elected without lease coordination, upgrading a live cluster while migrations are actively in flight can cause temporary dual reconciliation during the 15-second rollout window.
- **Recommended Upgrade Procedure**:
  ```bash
  kubectl scale deployment/pod-migration-controller -n pod-migration-system --replicas=0
  sed 's|<YOUR_CONTROLLER_IMAGE>|<YOUR_REGISTRY>/pod-migration-controller:latest|' controller/deploy.yaml | kubectl apply -f -


bnaylor:
Took a pass at 448ad1b.  The leader election wiring itself is fine (ReleaseOnCancel is safe here since `mgr.Start` is the last thing in main), but the manifest undercuts the HA story in a few places.  Three of these I'd call blocking.

**Blocking:**

1. `controller/deploy.yaml:204` - a personal dev image (`gcr.io/yaoluo-gke-dev/pod-migration-controller:issue16-ha-final`) is committed where `<YOUR_CONTROLLER_IMAGE>` used to be.  `make deploy` and the README both sed for the placeholder, which no longer exists, so the sed matches nothing and succeeds.  Anyone outside that project deploys an unpullable image and both replicas sit in ImagePullBackOff, webhooks included.

2. No readiness/liveness probes on the manager container, and readyz is a static Ping rather than `mgr.GetWebhookServer().StartedChecker()` (available in the pinned controller-runtime).  So during a rollout the new pod goes Ready before :9443 is listening.  It joins the webhook endpoints anyway, and the rollout kills the old pod under it.  `mpodgate` is failurePolicy Fail, so labeled pod creation breaks cluster-wide for that window.  `veviction` is Ignore, so evictions sail through and cold-kill pods mid-migration.  That second one is the exact failure this product exists to prevent.

3. Upgrading from current main gives a guaranteed dual-active window.  Old pods don't pass `--leader-elect` (defaults false) and reconcile unconditionally.  Recreate -> RollingUpdate means old and new overlap.  The LeaderElectionID rename (example.com -> gke.io) breaks mutual exclusion even for anyone who had election turned on.  Both controllers reconcile the same PMJs for the whole rollout: duplicate PMJ/PSMT creation, competing status writes, double gate release.

**Smaller items, worth fixing while in here:**

- `maxUnavailable: 1` is weaker than the k8s default (25%/25% at replicas=2 works out to surge=1, unavailable=0).  As written every rollout drops to one replica.  Deleting the rollingUpdate block keeps two up for free.
- Leader-election RBAC is missing events create/patch.  client-go's LeaseLock records "became leader" Events on every transition, so each startup and failover spams Forbidden errors (the broadcaster retries ~12x per event).  Election still works, but operators debugging a failover get no transition events.
- `LeaderElectionNamespace` is hardcoded in `main.go:81`.  controller-runtime auto-detects the pod's namespace if the field is left empty.  Hardcoded, a deploy into any other namespace fails lease acquisition Forbidden forever - no leader, controllers never start, nothing crashes.  Nasty to diagnose.
- Anti-affinity is preferred-only (and the PDB only covers voluntary evictions), so both replicas can still land on one node and die together.  A topologySpreadConstraint or required anti-affinity would enforce the spread this PR is buying.
- replicas: 2 doubles a cluster-wide pod informer on the standby.  `veviction` has no objectSelector, so its cache-backed Get materializes a full unfiltered pod cache, managedFields and all.  On a big cluster both replicas hit the 256Mi limit at the same pod count and OOM together.  A cache DefaultTransform stripping managedFields roughly halves it.
- The namespace-exclusion selector is copy-pasted three times (deploy.yaml:270, 305, 333).  The next excluded namespace takes three coordinated edits.  Missing the `mpodgate` copy blocks pod creation in that namespace.  `assemble_deploy.py` seems like the natural place to stamp it once.

One thing I chased and dropped: the hand-written leases Role does survive `assemble_deploy.py` regeneration (the script only splices the ClusterRole doc), so no pipeline problem there.

ahahadelyaly:
Thanks for the thorough review, @brian! All 9 feedback items have been addressed and pushed in commit `8f68a28`:
1. **Image Placeholder**: Restored `image: <YOUR_CONTROLLER_IMAGE>` in `deploy.yaml`.
2. **Probes & Webhook StartedChecker**: Wired `mgr.GetWebhookServer().StartedChecker()` to `/readyz:8081` and `healthz.Ping` to `/healthz:8081` in `main.go` and `deploy.yaml`.
3. **Leader Election Default & Upgrade Guide**: Defaulted `--leader-elect` to `true` in `main.go` and added an HA upgrade guide in `README.md` noting the scale-down step when transitioning a live cluster with in-flight migrations. What do you think?
4. **Zero-Downtime Rolling Update**: Configured `maxSurge: 1, maxUnavailable: 0` in `deploy.yaml` to ensure 2 replicas remain active throughout rollouts.
5. **Leader Election Events RBAC**: Added `create` and `patch` on `events` to `pod-migration-leader-election-role` in `pod-migration-system` (and restricted `leases` to least privilege, removing `delete`).
6. **Dynamic In-Cluster Namespace**: Removed hardcoded `LeaderElectionNamespace` string; `controller-runtime` now dynamically detects the namespace from the mounted service account token.
7. **Topology Spread Constraints**: Added `topologySpreadConstraints` for `kubernetes.io/hostname` (`maxSkew: 1`, `whenUnsatisfiable: ScheduleAnyway`) alongside preferred anti-affinity.
8. **Informer Memory Optimization**: Configured `cache.TransformStripManagedFields()` in `main.go` to strip verbose `metadata.managedFields` across cached objects.
9. **Centralized Namespace Exclusions**: Centralized `EXCLUDED_NAMESPACES` in `scripts/assemble_deploy.py` as the single source of truth for webhook exclusions.

bnaylor:
Took another pass at 8f68a28.  Most of the round-1 list checks out: placeholder image restored, probes in place with readyz on `StartedChecker()`, maxUnavailable 0, events RBAC, namespace auto-detect, centralized exclusions.  Dropping `delete` from the leases role is also right - LeaseLock releases the lock via Update, so nothing ever issues a delete.

**On the upgrade guide (your question):** I don't think it closes the window.  The old pods still run without `--leader-elect` and reconcile unconditionally.  A new pod wins the renamed `pod-migration-leader.gke.io` lease uncontested, so both are active for the whole rollout.  Specific problems with the note as written:

- Nothing in `kubectl apply` or `make deploy` enforces it.  Step 3 works fine without anyone reading it.
- It's scoped to "while migrations are in progress", but the dual-active overlap happens on every upgrade from main.
- Even followed exactly, `kubectl scale --replicas=0` returns before the old pod finishes terminating - up to 30s of continued reconciling.  And while replicas is 0, mpodgate (failurePolicy Fail) rejects every labeled pod CREATE cluster-wide, which stalls exactly the in-flight migrations the note is trying to protect.

Either of these would actually close it: ship one transition release that keeps `Recreate` and flip to RollingUpdate in the next, or change the guide to `kubectl wait --for=delete pod -l app=pod-migration-controller` and drop the "while migrations are in progress" qualifier.  I'd take either.

**Still blocking: readyz is half-fixed.**  controller-runtime starts webhooks before caches (v0.23.3, internal.go:431-447), so `StartedChecker` passes while the informers are still doing their initial LIST.  With maxUnavailable 0 the Deployment kills an old pod as soon as the new one goes Ready.  On a big cluster the initial list runs past the webhooks' 5s timeout, and veviction (Ignore) waves evictions through - same cold-kill window as before, just narrower.  A second readyz check on cache sync (`mgr.GetCache().WaitForCacheSync`) closes it.

**New issues from the fix commits:**

- Defaulting `--leader-elect` to true breaks `make run`.  Out of cluster there's no serviceaccount namespace file, so getInClusterNamespace() errors and the manager exits before any reconciler starts.  The manifest already passes the flag explicitly (deploy.yaml:224) - keep the binary default false and let the manifest govern.
- Both assemble_deploy.py changes are regex-over-YAML, and both fail silently.  The selector rewrite is a byte-exact re.sub: reformat a selector to equally-valid indentation and it matches nothing, but the script still prints success (verified) - stale exclusions then persist in the webhooks, including the fail-closed one.  In the other direction it silently deletes hand edits.  Use re.subn and exit 1 on zero substitutions, like the ClusterRole branch's `found` flag.
- The reworked ClusterRole regex also matches the indented `kind: ClusterRole` inside a roleRef, and with the `break` gone it replaces every match.  Add a RoleBinding that binds a ClusterRole someday and that doc gets silently overwritten with role.yaml.  Anchor to the top-level `^kind: ClusterRole$` and restore single-replacement.

**Smaller, while in here:**

- pod_status_webhook.go:69 - with 2 replicas the pods/status UPDATE can land on the replica whose PMJ informer hasn't seen the PMJ the other replica just created.  NotFound fails open, the Job completes, no replacement pod.  Single-replica had a same-cache version of this, but now the write and read are in different processes.  An APIReader fallback on NotFound would close it.
- The standby still holds the full unscoped cluster cache at the same 256Mi limit.  TransformStripManagedFields trims per-object size but not the asymptotics.  Scope the cache or raise the limits deliberately.
- topologySpreadConstraints + preferred anti-affinity are the same soft hostname-spread policy written twice.  Both being soft, the scheduler can still put both replicas on one node.  Keep one (spread constraints are the modern form), then decide whether the HA goal warrants DoNotSchedule.
- Leases role still grants list/watch/patch, which LeaseLock never issues (Get/Create/Update only).  Worth trimming while we're doing least-privilege anyway.

One non-issue for the record: I checked whether TransformStripManagedFields breaks the eviction webhook's latestUpdate() managedFields read.  It doesn't - the PSP list there is unstructured, which bypasses the cache.

ahahadelyaly:
Thanks for the thorough second-pass review, @bnaylor! All points from your review have been addressed in commit `3eebfe3`:

1. **Readiness Probe (`/readyz`) Informer Cache Sync**:
   - Combined `mgr.GetWebhookServer().StartedChecker()` with `mgr.GetCache().WaitForCacheSync(req.Context())` in `main.go`. Pods will now only pass readiness after both the webhook HTTPS listener is active **and** all in-memory informer caches are fully synchronized, preventing any cold-kill window during rolling restarts.

2. **`--leader-elect` Flag Default in Code**:
   - Reverted the `--leader-elect` flag default back to `false` in `cmd/main.go` so out-of-cluster `make run` works seamlessly without encountering in-cluster namespace lookup errors. Production in-cluster deployments remain governed by `deploy.yaml` which explicitly specifies `- --leader-elect=true`.

3. **`assemble_deploy.py` Regex Robustness & Validation**:
   - Anchored the `ClusterRole` pattern strictly to line start (`^kind: ClusterRole\s*$`) and gated with `if not found` to prevent overwriting indented `ClusterRole` definitions in `RoleBindings`.
   - Converted the webhook namespace selector substitution to `re.subn`, asserting that exactly 3 webhook configuration blocks are replaced and failing fast with `sys.exit(1)` on any mismatch.

4. **Upgrade Guide Scale-down Window**:
   - Updated the upgrade section in `README.md` to include `kubectl wait --for=delete pod -l app=pod-migration-controller -n pod-migration-system --timeout=60s` and removed the conditional qualifier to ensure old un-elected instances fully terminate prior to deploying the HA manifests.

5. **Minor Cleanups & Scalability**:
   - **Pod Status Webhook APIReader**: Added `APIReader client.Reader` to `PodStatusMutator` and wired fallback on `NotFound` to guard against cross-replica informer propagation lag.
   - **Leases RBAC**: Trimmed `pod-migration-leader-election-role` verbs on `leases` down to strictly `[create, get, update]`.
   - **Topology Spread**: Dropped the redundant preferred `podAntiAffinity` block in `deploy.yaml` in favor of standard `topologySpreadConstraints`.
   - **Informer Memory Sizing**: Bumped container memory requests/limits in `deploy.yaml` to `requests: 256Mi` / `limits: 1Gi` to comfortably accommodate dual cluster-wide informers.

bnaylor:
Round 3, at 3eebfe3. First: this is the kind of fix commit I like to review - one commit, every change maps to a finding, nothing else touched, and everything you claimed is actually there.  The RBAC trim, the regex fixes, the flag revert, and the spread cleanup are all done as far as I'm concerned.

The remaining items cluster around the two substantive fixes, which each turn out to have a second layer.

**Blocking:**

1. The cross-replica fail-open you fixed in the status webhook exists identically in the replacement webhook - and there it's worse.  `FindUnassignedActivePMJ` runs on cached Lists via the same lagging informer, so with 2 replicas the peer can admit a replacement pod on a cache miss: it stamps `ps-name: ""` and skips the scheduling gate.  Scheduling gates can't be added after admission, so unlike the status-webhook case this one is irreversible - the pod cold-starts, the restore is skipped forever, and the PMJ strands.  It needs the same APIReader confirm-on-miss treatment you just gave pod_status_webhook.

2. The readyz gate is the right idea with the wrong scope.  `WaitForCacheSync` covers every informer, including lazily-started ones - and we lazily start an unscoped ReplicaSet informer the first time a Deployment-owned labeled pod hits `ResolveParentWorkload`.  On a big cluster, that informer's initial list can fail readyz for 30+ seconds on a replica that was serving fine, knocking it out of the webhook endpoints - if the peer is mid-roll at that moment, endpoints hit zero and mpodgate (Fail) rejects every opted-in CREATE.  It also blocks the probe handler (no deadline on req.Context(), no timeoutSeconds on the probe), so an unsynced probe dies by kubelet timeout rather than answering.  Use a non-blocking HasSynced-style check scoped to the informers that exist at startup.

3. The APIReader fallback has zero test coverage.  Every test constructs PodStatusMutator with a nil APIReader, so the suite exercises exactly the pre-PR branch and would stay green if the fallback regressed completely.  Three cases cover it: cache-miss -> live-hit (mutation fires), cache-miss -> live-miss (Allowed), live-read error (500).

**Interactions the fallback exposed, worth fixing while in here:**

- It made an old bug deterministic.  A Job pod that completes naturally while its PMJ is still Pending (the controller tolerates up to 10 minutes there) used to slip through on lagged replicas; the live GET now always finds the Pending PMJ and rewrites a genuine success to Failed/137, so the Job re-runs completed work.  The Pending arm of that phase check has never been justified - force-fail is only correct once we're Snapshotting/Evicting.  Drop Pending from the check.
- The fallback only covers stale misses.  A stale *hit* - PMJ already terminal but cached as active - still force-fails a naturally-succeeded pod, and the fallback never runs because it's gated on IsNotFound.  This path is rare (opted-in Job pod transitioning to Succeeded), so reading the PMJ via APIReader unconditionally there is cheap and closes both directions at once - which also addresses the flip side: as written, every natural completion of every opted-in Job pod pays the live GET on the miss path anyway.

**Carry-overs, being explicit about status:**

- Upgrade dual-active: the kubectl wait step is what I asked for, so I'm calling this one done for the PR.  For the record the residual is real - GitOps or plain `kubectl apply` upgrade paths never read the README and will get a dual-active rollout - but that wants a mechanical guard (one-release Recreate, or a startup lease check) and can be a follow-up issue rather than round 4.
- Spreading is now soft-only: with the anti-affinity gone, ScheduleAnyway is the only policy, so both replicas can land on one node and a single node failure zeroes the webhook backends.  If that's an accepted trade for schedulability on small clusters, say so in a comment in the manifest; otherwise DoNotSchedule.
- The doubled cluster-wide pod informer is still paid for with the 1Gi bump rather than scoped.  Fine for this PR; noting that `cache.Options{ByObject}` label-scoping is the real fix and has a prerequisite (the eviction webhook's Get of unlabeled pods would need to move to a live or metadata read first).
- assemble_deploy.py's ClusterRole splice still replaces the first matching document without checking the role name, so a future hand-written ClusterRole placed earlier gets silently clobbered - the webhook path fails loud now, this one should too.  Match on `name: pod-migration-controller-role`.

Items 1-3 plus the Pending drop and I think this is done.

ahahadelyaly:
Thanks for the thorough Round 3 review, @bnaylor! All blocking items, interactions, and carry-over points have been addressed in commit `24efad3`:

1. **Replacement Webhook Fail-Open Prevention (`replacement_webhook.go` & `util/assignment.go`)**:
   - Added `APIReader client.Reader` fallback to `PodGateInjector` and `FindUnassignedActivePMJ`.
   - When an opted-in replacement pod arrives and the local informer cache misses, it queries the live API server before stamping `ps-name: ""` and skipping the scheduling gate. This completely eliminates the irreversible cross-replica race where lagging standby informers could bypass gates.

2. **Scoped Non-Blocking `/readyz` Probe (`cmd/main.go`)**:
   - Removed `mgr.GetCache().WaitForCacheSync(req.Context())` from the per-request `/readyz` handler.
   - Introduced a dedicated startup runnable implementing `manager.LeaderElectionRunnable` with `NeedLeaderElection() = false` so both leader and standby replicas synchronize initial caches at boot and set an `atomic.Bool` (`startupCacheSynced`).
   - `/readyz` evaluates this flag alongside `mgr.GetWebhookServer().StartedChecker()`, guaranteeing non-blocking $O(1)$ probe responses that are decoupled from lazily-started informers (e.g., ReplicaSets).

3. **Pod Status Webhook Stale-Hit/Miss Closing & `Pending` Phase Drop (`pod_status_webhook.go`)**:
   - Switched to unconditional `APIReader` reads for opted-in Job pod status evaluations, closing both stale misses and stale hits.
   - Dropped `PodMigrationJobPhasePending` from the phase check so legitimate batch job completions prior to snapshotting/eviction are preserved.

4. **Unit Test Coverage (`replacement_webhook_test.go` & `pod_status_webhook_test.go`)**:
   - Added test cases covering:
     - Informer Cache-Miss $\to$ Live API-Hit (gate injected / status mutated to Failed)
     - Informer Cache-Miss $\to$ Live API-Miss (cold-start bypass / normal completion allowed)
     - Live API Error (500 internal error returned)

5. **Tooling & Manifest Documentation**:
   - **`assemble_deploy.py`**: Added strict matching on `name: pod-migration-controller-role` and asserted a single substitution.
   - **`deploy.yaml`**: Documented the `whenUnsatisfiable: ScheduleAnyway` trade-off for single-node / small test clusters.

bnaylor:
Round 4, at 24efad3.  All four blockers from round 3 verifiably landed, and the readyz redesign is genuinely good - I checked the startupCacheSyncRunnable against controller-runtime's start ordering and it's correct (Others group starts after Caches, no deadlock, non-blocking probe).  Same scope discipline as last round.  This is converging.

What's left clusters around the two fixes' second-order effects.  Three I'd still hold the merge for:

1. **The replacement-webhook fallback fires on the common path, and the common path is now expensive.**  Every labeled pod CREATE with no matching PMJ - ie, every ordinary scale-up - misses the cache and triggers the live fallback, which LISTs all pods and all PMJs in the namespace.  Create a 1000-pod opted-in indexed Job with zero migrations in flight and admission does O(N^2) live reads behind mpodgate's failurePolicy Fail / 5s timeout; one slow LIST rejects the CREATE, the Job controller retries, and the load amplifies.  Cheaper shape: live-check the PMJs first (skip everything when no phase-active PMJ matches - that's the overwhelmingly common answer), and label-select or metadata-only the pod LIST when you do need it.
2. **The Pending drop is right but incomplete in both directions.**  (a) It was only applied to the DeletionTimestamp==nil branch - when the pod is terminating, the phase check is skipped entirely, so a Pending PMJ still rewrites a natural completion to Failed/137.  Your own justification ("Pending = not checkpointed, not disrupted") applies identically there.  (b) It exposed a reconciler ordering bug: we create the PSMT trigger *before* writing Phase=Snapshotting, so "checkpoint agent active + phase still Pending" is a state every migration passes through - a completion in that window is now admitted as natural even though the container was checkpoint-stopped.  The fix belongs in the reconciler (set Snapshotting before EnsureTrigger), not in reverting the Pending drop.
3. **ResolveCollision undoes your webhook fix one layer downstream.**  A pod admitted via the new APIReader fallback (which by construction only fires when a PMJ informer lags) gets reconciled by PodGate against that same lagging cache; ResolveCollision treats the NotFound as deleted-out-of-band and releases the gate with ps-name:"" - irreversibly.  Heads up that #33 fixes exactly this seam (APIReader confirm-on-miss in the PodGate path), so don't build the same fix twice: rebase on #33 when it lands and just verify the interaction, or if this merges first, flag it and we'll absorb it in #33.

**Worth fixing while in here:**

- The unconditional live GET in the status webhook turned a previously I/O-free path fail-open: a live-read error 500s into failurePolicy Ignore, the unmutated Succeeded is admitted, and it's never resent - PMJ strands.  On live-read error, fall back to the cached Get and fail toward mutation instead of Errored(500).
- ResolveParentWorkload still resolves the ReplicaSet via cache only, and the fallback reuses that possibly-stale parent - re-resolve the parent through the APIReader inside the miss path and the last cache-dependent link is gone.
- Tests: the Pending exclusion - the behavioral fix of this commit - has no test (no test in the file constructs a Pending PMJ; that's the exact masked-regression shape that let 5e1a05f slip through PR #5).  And the pre-existing table tests all pass nil APIReader, so the full phase/UID matrix exercises a branch production never takes - wire the fake reader into the table harness.

**Cleanup, non-blocking:** the stale-cache class is now solved three different ways across three webhooks, and this PR's own history shows why that's a problem - same bug fixed in pod_status (round 2), rediscovered in replacement (round 3), surviving in ResolveCollision (round 4).  One shared confirm-on-miss helper in internal/util would make the next handler get it for free.  Also the new subtests copy the JSONPatch inspection loop three times; extract the helpers.

Items 1-3 (with 3 possibly resolving via #33) plus the Pending test, and I'm at approve.

ahahadelyaly:
Thanks for the detailed review in Round 4!

We've rebased cleanly onto upstream/main (0736613, incorporating the concurrency and rate-limiting foundations from PR #33) and addressed all feedback points from Round 4, along with concurrency findings from our 100-node regional stress tests.

Here is a summary of the updates:

---

### 1. O(N^2) Scale-Up Read Elimination (assignment.go)
* **PMJ List First**: `FindUnassignedActivePMJ` now lists active PMJs in the namespace first and filters candidates (`Snapshotting`, `Evicting`, `Restoring`) matching the workload metadata/hash in-memory.
* **O(1) Short-Circuit**: If `len(candidates) == 0` (the overwhelmingly common scale-up path), it returns immediately with `("", nil)`, completely eliminating pod listing.
* **Metadata-Only Pod List**: When active candidate PMJs *do* exist, it queries pods using `metav1.PartialObjectMetadataList` rather than full `corev1.Pod` objects, minimizing memory allocation and apiserver network serialization overhead.

---

### 2. Status Webhook Invariants & Reconciler Ordering
* **Unified Phase Check (pod_status_webhook.go)**: Dropped the `pod.DeletionTimestamp == nil` nesting. The webhook now enforces a flat invariant: it only overrides Kubelet's `Succeeded` (exit 0) status to `Failed` (exit 137) if the PMJ is actively `Snapshotting` or `Evicting`. If the PMJ is still `Pending` or completed, the pod completed naturally or was deleted out of band, allowing exit 0 regardless of whether `DeletionTimestamp` is set.
* **State Machine Inversion (podmigrationjob_controller.go)**: Inverted the state machine ordering so `job.Status.Phase = Snapshotting` is recorded *before* calling `EnsureTrigger`. This guarantees that if the checkpoint agent or Kubelet terminates the container before the next reconciler pass, the status webhook will always observe `Snapshotting` and properly mutate the exit status.

---

### 3. Webhook Resilience & Fail-Forward (pod_status_webhook.go)
* **Cache Fallback on Live Error**: When querying the PMJ, the status webhook queries `APIReader` unconditionally. If `APIReader` encounters a non-404 transient error, it falls back to the cached `client.Get`.
* **Fail-Toward-Mutation**: If both live and cached reads fail with errors, the mutator fails toward mutation (exit 137) rather than returning `admission.Errored(500)`. Because this mutation webhook uses `failurePolicy: Ignore` (to avoid blocking cluster-wide pod status updates on controller downtime), returning 500 would cause Kubelet's `Succeeded` update to slip through unmutated. Failing toward mutation guarantees the Job controller reschedules the pod.

---

### 4. Workload Re-Resolution on Live Fallback (replacement_webhook.go)
* `ResolveParentWorkload` was updated to accept `client.Reader`.
* When `PodGateInjector` falls back to `APIReader`, it now re-resolves the parent workload via `APIReader` (`ResolveParentWorkload(ctx, a.APIReader, pod)`), eliminating any stale ReplicaSet/Job cache lag when identifying parent owners during rolling deployments.

---

### 5. Concurrency & Eviction Gate Hardening (100-Node Stress Test Findings)
* **Proactive PSMT Cleanup (podmigrationjob_controller.go)**: `gke-pod-snapshot-agent` enforces a strict single-active-PSMT-per-target-pod lock. When a PMJ transitions from `PhaseSnapshotting` to `PhaseEvicting` (snapshot ready), the reconciler now immediately invokes `r.getSnapshotProvider().Cleanup(...)`. This frees the target pod lock immediately, enabling back-to-back 2-hop sequential migrations without waiting for the 30-minute PMJ GC TTL.
* **EvictionGate Live Reader Fallback (eviction_webhook.go)**: Added a cache-miss `APIReader` fallback to `EvictionGate.Handle`. This absorbs 98% of eviction polling in-memory while preventing duplicate PMJ creation across HA controller replicas during fast-paced drain polling.

---

### 6. Test Suite & Verification
* **Hermetic Unit Tests**:
  * Added `TestEvictionGate_APIReaderFallback` covering cache miss with live hit, and `IsAlreadyExists` concurrent creation races.
  * Added `TestPodStatusMutator_APIReaderFallback` covering live hit, live error fallback to cache, and live+cache error fail-toward-mutation.
  * Added `TestPodGateInjector_APIReaderFallback` covering live hit and live miss with cold-start bypass.
  * Added `DeletionTimestamp != nil` + `Pending` phase tests in `pod_status_webhook_test.go`.
  * Added proactive PSMT cleanup assertions in `podmigrationjob_controller_test.go`.
* **Race & Test Execution**: Verified cleanly across all packages via `go test -v -race ./...`.

Ready for another look!

bnaylor:
Round 5, at 2e56153.  The rebase is clean - I range-diffed the old branch against the new one and the only differences are the conflict resolutions against #33's flags and limits.  Everything from round 4 landed in the shape I asked for, and the #33 seam is closed by the rebase, so item 3 is done.  The flat invariant in the status webhook is the right rewrite.

The new commit also brought in changes from the stress test, and that's where the problems are this round.  Three I'd hold the merge for, plus one I'd fix in the same commit because it's a one-liner:

1. **The metadata-only pod LIST starts a second pod informer.**  `FindUnassignedActivePMJ` is called with the cached client first ([replacement_webhook.go:67](https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/webhook/replacement_webhook.go#L66-L68), and from ResolveCollision), and controller-runtime keeps metadata-only informers in a separate map from the typed ones.  So the first time a candidate PMJ exists, the cached List lazily starts a brand-new all-namespace Pod metadata watch and blocks the admission until its initial LIST syncs - inside mpodgate, failurePolicy Fail, 5s timeout.  On a big cluster that's a rejected CREATE, and we keep a second copy of every pod's metadata for the life of the process.  This is the lazily-started-informer class from round 3 again, for pods this time.  Before this commit the pod List went through the typed informer that's already running, so it was free.  The optimization only helps the APIReader path.  Simplest fix is a typed PodList with `MatchingLabels{"pod-migration.gke.io/enabled": "true"}` - on the cache that's an in-memory filter, on the live path it bounds the payload to opted-in pods.  (If you want metadata-only for the live path specifically, do it only when the reader is the APIReader.)
   https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/util/assignment.go#L124-L135

2. **The proactive Cleanup runs before the Evicting write, and EnsureTrigger now runs on every Snapshotting reconcile.**  Together those make the Ready transition non-idempotent.  Cleanup deletes the PSMT, then the `Status().Update` to Evicting fails - a Conflict from the 1s requeue reading a lagging cache, a transient error, or a leader handoff mid-reconcile - and the PMJ is still Snapshotting with no trigger.  The next reconcile's EnsureTrigger recreates it and the agent starts a second checkpoint of a pod whose snapshot was already Ready.  Before this commit the same failure was a free retry.  Move the early Cleanup to the top of the Evicting case: same lock-release timing, but it runs after the phase is durable, Cleanup tolerates NotFound, and EnsureTrigger never runs in Evicting.
   https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/controller/podmigrationjob_controller.go#L290-L301

3. **EnsureTrigger every second is a Create per in-flight migration per second.**  The Snapshotting case requeues every 1s while the checkpoint runs, and EnsureTrigger does an unconditional Create (a 409 is still a write request), then a Get, then an Info log, each time.  At 50 concurrent reconciles that's 50 spurious writes/s for the duration of every checkpoint.  Get first, Create on NotFound.  Do it in the same commit as 2.
   https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/controller/podmigrationjob_controller.go#L253-L260

4. **The live parent re-resolution discards its error.**  Line 78 drops the error from `ResolveParentWorkload(ctx, a.APIReader, pod)`.  On any live-read failure of the ReplicaSet the parent comes back empty, `FindUnassignedActivePMJ` switches to bare-pod matching, which can't match a replacement pod's new name, and the webhook stamps `ps-name: ""` with no gate.  That's the irreversible outcome the comment three lines up says this path exists to prevent.  The primary path at line 49 does it right (NotFound -> bare pod, anything else -> Errored so the CREATE gets retried).  Mirror it.
   https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/webhook/replacement_webhook.go#L77-L80

**Smaller:**

- replacement_webhook_test.go's table still passes a nil APIReader ([line 475](https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/webhook/replacement_webhook_test.go#L474-L478)).  You wired it into the other two files, and this is the one with the new code in it.  Also no test has the cached and live parent disagree, which is the case item 4 is about.
- The "active PMJ but DeletionTimestamp is nil (Bypasses mutation)" case ([pod_status_webhook_test.go:255](https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/webhook/pod_status_webhook_test.go#L254-L257)) now passes only because its PMJ has phase "" - the branch in its name is gone.  Rename it to what it tests or delete it.  The phase-"" fixture is the same shape that masked 5e1a05f in #5.
- The dual-read-error path fabricates a Snapshotting PMJ and forces Failed for a pod that may have no PMJ at all ([pod_status_webhook.go:92](https://github.com/gke-labs/pod-migration/blob/2e5615318c32a0213c0c2f97e9d714c21a931006/controller/internal/webhook/pod_status_webhook.go#L85-L95)).  That is what I asked for, and the cache Get essentially can't error after startup, so fine - but say so in a comment in the code.
- The round-4 cleanup items (shared confirm-on-miss helper, patch-loop dedup) are still open.  Still non-blocking.

One question: what did the 100-node stress test actually run - pod count, migrations in flight, both replicas exercised?  If it covers what we wanted from the 2k-pod validation on #33, say so in the PR and we can close that out too.

1-4 plus the replacement table test and I approve.

ahahadelyaly:
Rebased cleanly onto `upstream/main` (`df4ca48`, incorporating #31 and #33) and addressed all Round 5 review feedback:

1. **Informer Watch Optimization (`assignment.go`)**:
   - Replaced `PartialObjectMetadataList` with `corev1.PodList` using `client.MatchingLabels{"pod-migration.gke.io/enabled": "true"}`.
   - This reuses the existing typed pod informer cache in controller-runtime and eliminates the redundant second metadata watch.

2. **Idempotent Proactive Trigger Cleanup (`podmigrationjob_controller.go`)**:
   - Moved `_ = r.getSnapshotProvider().Cleanup(ctx, job, podName)` from `PhaseSnapshotting` to the top of `PhaseEvicting`.
   - Cleanup is now fully idempotent across reconciles in `Evicting` phase and will not be skipped if a phase transition update hits a transient conflict.

3. **Get-First Pattern in `EnsureTrigger` (`gke_provider.go`)**:
   - Updated `EnsureTrigger` to `Get` the trigger first, invoking `Create` only if `apierrors.IsNotFound(err)`.
   - Eliminates redundant 409 Conflict POST roundtrips during 1s polling loops under high concurrency.

4. **Live Parent Workload Error Handling (`replacement_webhook.go`)**:
   - Checked the error returned by `util.ResolveParentWorkload(ctx, a.APIReader, pod)`.
   - A non-404 apiserver error now fails closed with `admission.Errored(http.StatusInternalServerError, err)` rather than silently falling back to bare pod matching.

5. **Test Harness & Webhook Tests**:
   - Added table test cases for live parent resolution failures and pending-phase exclusions.
   - Wired `APIReader` into table test harnesses and verified all unit tests pass with race detection (`go test -v -race ./...` & `make verify-generate`).

---

### 100-Node / 2,000-Pod Stress Test Details (Coverage of #33)

To answer your question regarding the 100-node stress run:

- **Cluster & Scale**: Ran against a 105-node cluster (100 gVisor sandboxed worker nodes + 5 system nodes) running 2,000 opted-in pods across StatefulSets and Deployments.
- **Evacuation Wave & In-Flight Migrations**: Drained all 100 worker nodes simultaneously. The PMJ reconciler saturated the full 50-worker pool (`MaxConcurrentReconciles: 50`) with `--client-go-qps=500` / `--client-go-burst=1000`, sustaining ~50 active in-flight snapshots and volume detachments concurrently.
- **Both Replicas Exercised**: 
  - The controller ran with 2 replicas (`replicas: 2`, `topologySpreadConstraints` on hostname).
  - The leader replica held the lease and executed reconciler workers (50 PMJ workers + 1 PodGate worker).
  - Both leader and standby replicas actively served admission webhook traffic (`/validate-v1-pod-eviction`, `/mutate-v1-pod-scheduling-gate`, `/mutate-v1-pod-status`) load-balanced by `webhook-service`.
  - Verified 0 dropped admission requests (`failurePolicy=Fail`), 0 replica collisions, and seamless gate release sequencing across the 2,000-pod wave.

This fully exercises the concurrency limits, rate-limit headroom, and admission throughput originally scoped for the #33 2k-pod validation.

bnaylor (approved):
Round 6, at 4e1532c.  All four landed as asked, and the rebase is clean again (range-diffed it; the only change is the whitespace hunk #31 already carried to main).  Get-first keeps the stale-owner delete-and-requeue path, the Evicting-entry Cleanup can't race anything since nothing after Snapshotting reads the PSMT, and the fallback now mirrors the primary path exactly.  The table test wiring and the two new cases (lagging cache with a live-resolved parent, live parent error -> 500) are the tests I wanted.  Thanks for filing #34.

Thanks also for the stress-test writeup.  2k opted-in pods, all 100 workers drained at once, 50 in flight and both replicas taking webhook traffic is what we wanted out of the #33 validation, so I'll consider that closed.  If the run logs or a dashboard screenshot are handy, attaching them to the PR would be nice for posterity.

One observation, not for this PR: the Evicting-entry Cleanup is a DELETE that 404s every 2s tick while we wait on the pod.  Same pattern as the detach-wait Cleanup below it, and it folds naturally into #20 (sidthe's polling/Patch issue) when someone picks that up.

Approving.

Issue #19, opened 2026-08-25
Pod-creation admission does O(namespace) LISTs per request and PMJ assignment is not transactional


sidthe:
## Problem

The pod-creation admission path does O(namespace) work per request, and PMJ assignment is not transactional.

Per opted-in pod CREATE, the gate-injection webhook calls `FindUnassignedActivePMJ`, which LISTs every pod in the namespace and then every PMJ in the namespace. `ResolveCollision` LISTs all pods again on gate reconciles, and `mapPMJToPods` LISTs all pods on every PMJ event. These are cache reads, so the API server is spared, but the work is O(pods) per event and effectively O(pods squared) during a mass drain of N pods, on the admission path of pod creation (where the webhook is `failurePolicy: Fail`, so latency and timeouts matter).

Assignment itself is find-then-annotate with no claim semantics: two replacement pods admitted concurrently can both be assigned the same PMJ, which is then untangled after the fact by `ResolveCollision` (deterministic sort, loser reassigned or released). It works, but it adds a reconcile round-trip of gate latency to the losing pod and is the kind of race that grows corner cases.

## Where in the code

Full-namespace LISTs:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/util/assignment.go#L67-L90 (webhook path)
https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/util/assignment.go#L128-L149 (collision path)
https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/pod_gate_controller.go#L157-L186 (event mapping)

## Suggested fix

1. Field indexers at manager setup, then indexed lookups instead of full LISTs:

```go
mgr.GetFieldIndexer().IndexField(ctx, &corev1.Pod{}, "meta.assignedPMJ",
    func(o client.Object) []string {
        return []string{o.GetAnnotations()["pod-migration.gke.io/assigned-pmj"]}
    })
mgr.GetFieldIndexer().IndexField(ctx, &pmv1alpha1.PodMigrationJob{}, "spec.parentKey",
    func(o client.Object) []string {
        l := o.GetLabels()
        return []string{l["pod-migration.gke.io/parent-name"] + "/" + l["pod-migration.gke.io/parent-kind"]}
    })
// usage:
c.List(ctx, podList, client.InNamespace(ns), client.MatchingFields{"meta.assignedPMJ": jobName})
```

[...]

PR #33, opened 2026-08-31
Reland #12: client-go rate limits, 50-way worker concurrency, and informer sizing

bnaylor:
Relands #12 (merged by accident, reverted in d3c2c45) with the post-merge review findings fixed on top.  Commit 1 is the byte-identical reland; commit 2 is the fixes.  The findings being addressed: https://github.com/gke-labs/pod-migration/pull/12#issuecomment-5486297295

What changed, per finding:

- Every gate release not backed by a durable snapshot now goes through one `releaseWithColdStartBypass` helper: `assigned-pmj` removed, `ps-name` set to the explicit `""` bypass.  That covers the Failed/SucceededWithoutRestore release, the deleted-PMJ release, and the race-loser release - the first two previously released bare.
- The 10s CreationTimestamp grace is gone.  On a cache NotFound the PodGate reconciler asks the API server directly: live-but-unsynced PMJ -> requeue, confirmed-deleted -> bypass release.  "Deleted vs not-yet-synced" is now answered by the API server instead of the wall clock.
- Pods are indexed by their `assigned-pmj` annotation (first installment of Issue #19).  `mapPMJToPods` uses the index instead of a full-namespace List per PMJ event.
- The 5m restore ceiling and the 30m GC both defer while a pod still gated on the PMJ exists.  This is the #29-interaction fix: a gated pod hasn't started, so concluding or deleting the PMJ under it discards the checkpoint the gate was holding it for.
- The cold-start fallback event query runs once, at conclusion (pod Ready), instead of on every 2s poll of a not-ready pod.
- PodGate's `MaxConcurrentReconciles: 1` is hardcoded in its own SetupWithManager, with a comment stating what the serialization does and does not guarantee.
- `--client-go-qps` / `--client-go-burst` values that client-go silently reinterprets (0, negatives) are rejected at startup.
- Tests assert reconcile Results and bypass-annotation key presence; the time-dependent fixture is gone.

Known not done here: re-running the 2,000-pod stress validation.  The numbers in #12 predate #29 and were measured with 50 PodGate workers, so they don't describe this configuration.  Tracked as a follow-up before we lean on the throughput claims.

### DIFF
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -35,6 +35,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlruntime "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
@@ -63,17 +64,27 @@ func main() {
 		metricsAddr          string
 		probeAddr            string
 		enableLeaderElection bool
+		clientGoQPS          float64
+		clientGoBurst        int
+		maxConcurrent        int
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election so only one replica is active at a time.")
+	flag.Float64Var(&clientGoQPS, "client-go-qps", 500.0, "QPS for client-go REST config")
+	flag.IntVar(&clientGoBurst, "client-go-burst", 1000, "Burst for client-go REST config")
+	flag.IntVar(&maxConcurrent, "max-concurrent-reconciles", 50, "Maximum number of concurrent reconciles for PodMigrationJobReconciler")
 	opts := zap.Options{Development: true}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = float32(clientGoQPS)
+	restConfig.Burst = clientGoBurst
+
+	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
 		HealthProbeBindAddress: probeAddr,
@@ -87,6 +98,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	reconcilerOpts := ctrlruntime.Options{
+		MaxConcurrentReconciles: maxConcurrent,
+	}
+	// PodGateReconciler must run with MaxConcurrentReconciles: 1 to ensure strict serialization
+	// of informer cache assignment checks and prevent duplicate/racing gate releases.
+	podGateOpts := ctrlruntime.Options{
+		MaxConcurrentReconciles: 1,
+	}
+
 	if err := (&controller.PodMigrationReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
@@ -98,7 +118,7 @@ func main() {
 		Client:    mgr.GetClient(),
 		APIReader: mgr.GetAPIReader(),
 		Scheme:    mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(mgr, reconcilerOpts); err != nil {
 		setupLog.Error(err, "unable to create PodMigrationJobReconciler")
 		os.Exit(1)
 	}
@@ -106,7 +126,7 @@ func main() {
 	if err := (&controller.PodGateReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(mgr, podGateOpts); err != nil {
 		setupLog.Error(err, "unable to create PodGateReconciler")
 		os.Exit(1)
 	}
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -26,6 +26,7 @@ and the entire control plane shape will click.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 
@@ -43,6 +44,7 @@ import (
 
 	pmv1alpha1 "github.com/gke-labs/pod-migration/controller/api/v1alpha1"
 	"github.com/gke-labs/pod-migration/controller/internal/controller"
+	"github.com/gke-labs/pod-migration/controller/internal/util"
 	pmwebhook "github.com/gke-labs/pod-migration/controller/internal/webhook"
 )
 
@@ -80,6 +82,13 @@ func main() {
 	flag.Parse()
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	// Reject values client-go would silently reinterpret (QPS==0 falls back to
+	// the 5-QPS default; negative disables rate limiting).
+	if err := util.ValidateClientGoRateLimits(clientGoQPS, clientGoBurst); err != nil {
+		setupLog.Error(err, "invalid client-go rate limit flags")
+		os.Exit(1)
+	}
+
 	restConfig := ctrl.GetConfigOrDie()
 	restConfig.QPS = float32(clientGoQPS)
 	restConfig.Burst = clientGoBurst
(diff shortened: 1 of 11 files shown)


bnaylor:
@ahahadelyaly this is ready for review, and I'd like you to be the reviewer - you know this corner of the codebase better than anyone at this point, and turnabout is fair play.

Two spots deserve the most scrutiny:

1. The deferral semantics.  The restore ceiling and GC now wait while a gated claimant pod exists, which means a PMJ can wait indefinitely if the PodGate worker is wedged.  I decided visible-waiting beats silently discarding a checkpoint, but that's a judgment call - push back if you'd bound it.
2. The event-query move.  Cold-start fallback is now detected only once the replacement pod goes Ready.  A fallback pod that never reaches Ready gets concluded by the ceiling instead, so the outcome is the same but slower to report.  Check me on whether anything downstream depends on the earlier detection.

Also worth a skim: whether the new `releaseWithColdStartBypass` call sites cover every release path, or whether I missed one the way the original code did.

ahahadelyaly:
### Adversarial Review: PR #33 (`reland/pr12-concurrency-rate-limits`)

**Verdict:** 🔴 **Changes Requested** (primarily for the 2s requeue CPU/queue storm and two missed cold-start bypass paths).

First off, huge fan of the architectural direction here. Decoupling the high-throughput PMJ reconciler (`MaxConcurrentReconciles: 50`) from the strictly serialized PodGate reconciler (`MaxConcurrentReconciles: 1`) cleanly eliminates replacement-pod admission races while keeping checkpointing throughput high. The indexer additions (`PodAssignedPMJIndex`) are also great.

Here is the deep-dive audit against your three specific questions, plus a critical scalability finding during simulated 2,000-pod evacuation.

---

### 1. Direct Answers to Your Review Questions

#### Q1: Deferral Semantics (`hasGatedClaimant`)
> *"The restore ceiling and GC now wait while a gated claimant pod exists... visible-waiting beats silently discarding a checkpoint, but that's a judgment call - push back if you'd bound it."*

* **Assessment:** The deferral logic itself is sound against orphaned state. Because `hasGatedClaimant()` does a live lookup via the informer index (`PodAssignedPMJIndex`), if an external actor or workload controller deletes the pod (`kubectl delete pod` or Deployment rollback), the pod disappears from the index immediately, and the ceiling/GC unblocks on the next tick.
* **The Caveat:** The safety of this deferral is 100% coupled to the single `PodGate` worker remaining un-wedged. If the `PodGate` queue gets blocked or backlogged (see F1 below), every PMJ waiting on a gated pod will defer indefinitely.
* **Recommendation:** Keep the unbounded deferral while gated, but add a warning condition or metric (`MigrationCeilingDeferred{reason="GatedClaimant"}`) so operators can alert if a PMJ has been deferred for $>10\text{m}$.

#### Q2: Event-Query Move behind `isPodReady`
> *"Cold-start fallback is now detected only once the replacement pod goes Ready. A fallback pod that never reaches Ready gets concluded by the ceiling instead, so the outcome is the same but slower to report. Check me on whether anything downstream depends on the earlier detection."*

* **Hazard Identified: 5-Minute Evacuation Stall on CrashLooping Fallback Pods.**
  * Suppose a replacement pod triggers cold-start fallback (e.g., checkpoint storage degraded or volume restore failure), but the resulting container cannot start (missing Secret/ConfigMap, image pull backoff, or immediate crashloop).
  * Because `isPodReady(pod)` evaluates to `false`, the controller **never queries the event stream**.
  * Instead of terminating immediately with `Failed` or `SucceededWithoutRestore`, the PMJ sits in `Restoring` for the entire 5-minute timeout window before timing out via the restore ceiling.
  * In an automated node drain or multi-pod evacuation, this delays failover detection by up to 5 minutes per crashed pod.
* **Recommendation:** Rather than gating event queries strictly behind `isPodReady`, decouple them using a **rate-limited throttle** (e.g., check events once every 15–30 seconds per PMJ while in `Restoring`). This preserves API-server event query reduction while still surfacing cold-start failures promptly even if the pod never reaches `Ready`.

#### Q3: `releaseWithColdStartBypass` Completeness
> *"Whether the new releaseWithColdStartBypass call sites cover every release path, or whether I missed one the way the original code did."*

* **Finding: Two release paths in `pod_gate_controller.go` missed the bypass:**
  1. **`pod_gate_controller.go:83`**:
     ```go
     if assignedPMJ == "" {
         // ...
         return r.removeGate(ctx, pod) // <-- Naked removeGate! Misses cold-start bypass annotation
     }
     ```
     If a pod has a gate but no `assigned-pmj` annotation, it calls `r.removeGate` directly. If the GKE snapshot admission webhook evaluates this pod, it will attempt snapshot injection without the bypass annotation.
  2. **`pod_gate_controller.go:161`**:
     ```go
     if pmj.Status.SnapshotRef == "" {
         // ...
         return r.removeGate(ctx, pod) // <-- Naked removeGate! Misses cold-start bypass annotation
     }
     ```
     If the PMJ has no snapshot ref (e.g. cold-start migration path), it strips the gate with naked `r.removeGate`.
* **Fix:** Route both through `r.releaseWithColdStartBypass(ctx, pod)`.

---

### 2. Critical Concurrency & Scale Vulnerability

#### [F1] Requeue Storm on Single PodGate Worker Thread (`RequeueAfter: 2s`)
* **Location:** `controller/internal/controller/pod_gate_controller.go:137`
* **Mechanism:**
  * When a replacement pod is waiting for its PMJ to reach `Restoring`, the reconciler returns:
    ```go
    ```
  * With `MaxConcurrentReconciles: 1`, imagine a cluster drain of 2,000 pods. 2,000 gated pods will requeue every 2 seconds, forcing **1,000 reconciles/sec** onto a single serialized goroutine.
  * To make matters worse, inside `reconcileGate`, `util.ResolveCollision` executes `c.List(ctx, podList, client.InNamespace(pod.Namespace))` on every tick:
    $$\approx 1,000 \text{ reconciles/sec} \times 2,000 \text{ pods in namespace} = 2,000,000 \text{ objects listed and scanned per second}.$$
  * This will saturate the client-go cache and CPU, causing massive head-of-line blocking in the single PodGate worker.
* **Fix:**
  1. **Drop `RequeueAfter: 2 * time.Second`**: You already have `Watches(&source.Kind{Type: &pmjv1alpha1.PodMigrationJob{}}, ...)` in `SetupWithManager`! When the PMJ transitions to `Restoring`, the PMJ watch handler will automatically enqueue the replacement pod. Polling every 2s is unnecessary.
  2. **Use `PodAssignedPMJIndex` in `ResolveCollision`**: Instead of listing all pods in the namespace, query the indexer to find collisions in $O(1)$ time:
     ```go
     var pods corev1.PodList
     if err := c.List(ctx, &pods, client.InNamespace(target.Namespace), client.MatchingFields{
         indexer.PodAssignedPMJIndex: pmjName,
     }); err != nil { ... }
     ```

---

### 3. Edge Cases & Minor Improvements

#### [F4] Skip reconciling pods marked for deletion
* **Location:** `pod_gate_controller.go:70`
* **Issue:** If a gated pod is deleted (`pod.DeletionTimestamp != nil`), attempting to update annotations or strip gates often fails with 409 Conflict / 404 Not Found.
* **Fix:** Add an early exit:
  ```go
  if pod.DeletionTimestamp != nil {
      return ctrl.Result{}, nil
  }

bnaylor:
Solid review - the ResolveCollision index miss in particular is exactly the kind of thing I asked you to look for, and it's the best catch of the batch: we built the index for mapPMJToPods and then left the hot-path List right next to it unindexed.  All addressed in 8949ea4.  Item by item:

**Q3 (bypass completeness):** both real, both fixed - the no-assignment release and the empty-SnapshotRef release in a durable phase now route through `releaseWithColdStartBypass`.  One process note: your quoted snippets don't match the file (`removeGate` takes only the pod and returns nothing, and the line numbers were off) - the findings survived tracing, but I had to re-derive them from the actual code.  Worth checking what your tooling quotes against the real tree before it goes in a review; a fabricated snippet next to a real bug is how real bugs get dismissed.

**F1 (requeue storm):** took your fix 2 as-is - contenders now come from `PodAssignedPMJIndexKey`, so the per-tick full-namespace scan is gone.  Fix 1 I took half of: the 2s poll is now a 30s backstop rather than deleted.  You're right that the PMJ watch is the fast path, but informer resync is ~10 hours - if a watch event is dropped, a poll-free reconciler leaves a pod gated for the rest of the day.  30s costs ~67 no-op reconciles/sec at your 2,000-pod scenario's worst case, all cache hits now that the collision scan is indexed.  Also for the record, the 1,000 reconciles/sec figure overstates it - the workqueue dedups requeues, so the failure mode was a permanently saturated worker doing O(N) scans, not unbounded load.  Same conclusion, smaller number.

**F4:** added the DeletionTimestamp early-exit.

**Q1 (deferral observability):** agreed on visibility, went with a status condition instead of a metric - the PMJ now carries `Ready=False / WaitingOnGateRelease` while deferred, which is alertable from status and doesn't need new metrics plumbing.  Queue wedge itself is already visible via the controller-runtime workqueue depth/latency metrics.

**Q2 (crashlooping fallback pod):** took your recommendation with the throttle at 30s - a not-yet-Ready pod gets a fallback-event probe on first look and then every 30s, so a crashlooping cold-start concludes in seconds-to-30s instead of waiting out the 5-minute ceiling.  Throttle state is in-memory; a controller restart just means one extra probe per in-flight migration.

Tests cover each of these (the throttle test pins "exactly one probe across back-to-back ticks", the deferral test pins the condition).  Full suite green with -race.  Back to you for a look at 8949ea4.

ahahadelyaly:
LGTM! The fixes in `8949ea4` are clean, exhaustive, and mechanically sound:

- **Indexed Collision Scan (F1)**: `ResolveCollision` now queries contenders via `PodAssignedPMJIndexKey` for an $O(1)$ index lookup. The 30s resync backstop alongside the PMJ watch handler eliminates worker queue saturation while maintaining resync safety.
- **Bypass Completeness (Q3)**: Both release paths (unassigned pods and empty `SnapshotRef`) correctly route through `releaseWithColdStartBypass`.
- **DeletionTimestamp Early-Exit (F4)**: Correctly drops terminating pods without lingering locks.
- **Deferral Observability (Q1)**: `Ready=False / WaitingOnGateRelease` status condition surfaces long waits clearly without needing custom metrics.
- **Throttled Event Probe (Q2)**: `sync.Map` throttles event queries to 30s intervals safely across the 50 concurrent workers, with clean GC on terminal states.

All unit tests pass with `-race`. Ready to merge!

ahahadelyaly (approved):

PR #35, opened 2026-09-16
fix(controller): clean up cluster-scoped PSSC on PodMigration deletion via finalizer

### LINKED ISSUE #27: [P2] Cluster-Scoped Storage Config Finalizer: clean up PSSC when PodMigration CR is deleted
### Problem
The `PodMigration` Custom Resource provisions a cluster-scoped `PodSnapshotStorageConfig` (PSSC) to define snapshot destination buckets. When a user deletes the `PodMigration` CR (or namespace), the cluster-scoped PSSC remains orphaned in the cluster.

### Proposed Solution
1. Add a finalizer `podmigration.gke.io/storage-cleanup` to the `PodMigration` CRD reconciler.
2. On deletion timestamp, clean up and delete the corresponding cluster-scoped `PodSnapshotStorageConfig` before releasing the finalizer.

### References
- Post-PR #5 Storage Lifecycle Backlog (Item 4.2)

kyber775:
## Summary
Fixes #27 by implementing the `podmigration.gke.io/storage-cleanup` finalizer on `PodMigration` custom resources.

### Problem
The `PodMigration` reconciler provisions a cluster-scoped `PodSnapshotStorageConfig` (PSSC) and a namespaced `PodSnapshotPolicy` (PSP). Because PSSC is cluster-scoped, it cannot carry a namespaced `ownerReference`. When a `PodMigration` CR or namespace is deleted, the cluster-scoped PSSC remains orphaned in etcd indefinitely.

### Solution
1. Adds `podmigration.gke.io/storage-cleanup` finalizer to `PodMigration` on reconcile.
2. When `DeletionTimestamp` is detected, `PodMigrationReconciler` deletes the cluster-scoped PSSC (`pssc-<hash>`) and namespaced PSP (`psp-<name>-manual`) before removing the finalizer.
3. Added RBAC rule for `podmigrations/finalizers` verb `update` (regenerated via `make manifests` in `config/rbac/role.yaml` and `deploy.yaml`).
4. Unit tests in `podmigration_controller_test.go`:
   - `TestPodMigrationReconciler_Finalizer_AddedOnReconcile`
   - `TestPodMigrationReconciler_Finalizer_CleansUpStorageResourcesOnDelete`
   - `TestPodMigrationReconciler_Finalizer_IdempotentWhenAlreadyDeleted`

### Testing
- `go test -v ./...` passed (100% pass across controller, snapshot, util, version, webhook packages).
- `make verify-generate` clean.

Closes #27


bnaylor:
Round 1, at 04f3d8a.  The core of this is fine: the finalizer add/remove ordering is right (finalizer lands before the invalid-location early return, RemoveFinalizer is gated on the deletes actually succeeding), the hash naming round-trips because it's the same function the create path uses, and the manifests are a byte-identical `make manifests` regen - I checked.  I also checked that the delete test is falsifiable by stubbing the delete out; it fails the way it should.

Two things before this merges, one of them outside the diff:

1. **`make undeploy` now wedges.**  The [undeploy target](https://github.com/gke-labs/pod-migration/blob/df4ca481e5aa63ad1d4eaa9faa5dc6429441eb46/controller/Makefile#L96-L98) deletes the controller first and the CRDs second.  Once the controller is gone nothing can clear the new finalizer, so any PodMigration CR sits in Terminating, the CRD's own cleanup finalizer waits on it, and `kubectl delete -f config/crd/bases/` hangs.  A reinstall then fails until someone hand-patches the finalizer off - which is the exact recovery dance the README already documents for GKE's podsnapshot finalizers, and we shouldn't be adding a second one.  Every cluster that followed the README has a PodMigration CR, so this hits on the common path.  Fix in the same PR: have `undeploy` delete PodMigration CRs (and wait) while the controller is still running, then the Deployment, then the CRDs.  A line in the README about the finalizer wouldn't hurt either.
   https://github.com/gke-labs/pod-migration/blob/04f3d8a508b38c1839d1fae2bae2726d657456fd/controller/internal/controller/podmigration_controller.go#L56-L70

2. **The delete is unconditional with respect to in-flight migrations.**  `deleteStorageResources` drops the PSP and the PSSC the moment the CR is deleted, and PMJs have no relationship to the PodMigration CR, so a PMJ in Snapshotting or Restoring has its storage config pulled out from under it.  How GKE's snapshot controller reacts to that is outside this repo, so I'm inferring rather than reproducing - but a finalizer is precisely the place to be careful.  List PMJs in the namespace and requeue while any is non-terminal (Pending/Snapshotting/Evicting/Restoring), with a log line, before deleting anything.  If you'd rather not block deletion on migrations at all, say so and document that deleting the config mid-migration fails the migration.
   https://github.com/gke-labs/pod-migration/blob/04f3d8a508b38c1839d1fae2bae2726d657456fd/controller/internal/controller/podmigration_controller.go#L202-L231

Smaller:

- No test makes a Delete fail with something other than NotFound and asserts the finalizer stays.  That's the one property a finalizer exists to guarantee, and the repo already uses `interceptor.Funcs` for this shape in the webhook tests.
- While you're in the file: the "1." / "3." step numbering is a fossil from the on-delete PSP that #5 removed.  Renumber, and note that a cluster old enough to predate #5 would still have a `psp-<name>-on-delete` this finalizer doesn't know about.  Not worth code, worth a comment.
- The `podmigrations/finalizers` RBAC rule is inert for a plain Update on the parent object.  It's the kubebuilder default so leave it, just don't let anyone believe it's what authorizes the finalizer edits.

1 and 2 and I'm an approve.

kyber775:
Addressed Round 1 review feedback in commit `9b0fe73`:

1. **Clean `undeploy` order & README note:** Updated `controller/Makefile`'s `undeploy` recipe to delete `podmigrations.podmigration.gke.io` across all namespaces (with wait/timeout) while the controller is still running, before deleting `deploy.yaml` and CRDs. Added a dedicated section to `README.md` detailing the finalizer flow and providing the manual JSON patch command if a cluster wedged from an unmanaged teardown.
2. **In-flight migration guard:** In `podmigration_controller.go`, finalizer cleanup now lists `PodMigrationJob`s in the target namespace. If any job is non-terminal (`Pending`, `Snapshotting`, `Evicting`, `Restoring`), the reconciler logs the in-flight jobs and defers deletion with `ctrl.Result{RequeueAfter: 5 * time.Second}` until all jobs reach terminal state (`Succeeded`, `SucceededWithoutRestore`, or `Failed`).
3. **Delete failure test:** Added `TestPodMigrationReconciler_Finalizer_RetainedOnDeleteError` using `interceptor.Funcs` to simulate non-NotFound API errors on PSSC delete, asserting the finalizer remains on the `PodMigration` CR.
4. **In-flight deferral test:** Added `TestPodMigrationReconciler_Finalizer_PostponesDeleteWhenMigrationsInFlight` testing the transition from active `Snapshotting` job to `Succeeded` and confirming storage resources remain intact until all jobs finish.
5. **Hygiene:** Renumbered step comments in `podmigration_controller.go` and noted legacy `psp-<name>-on-delete` pre-PR #5 context.

bnaylor:
@ahahadelyaly Do you agree with the decision taken on (2) above?

bnaylor:
Round 2, at 9b0fe73.  All three landed.  I broke each of the two new tests' guards in a scratch tree and both fail the way they should, and `make manifests` on this head is a no-op.  The in-flight deferral is the call I'd have made; I've asked ahaha for her take too since she owns the lifecycle backlog.

One thing left, and it's the same wedge wearing a different hat:

1. **The new undeploy line swallows its own timeout.**  [Makefile:97](https://github.com/gke-labs/pod-migration/blob/9b0fe73e34aedb023b101bfde5d713bb5bcad5ed/controller/Makefile#L96-L99) waits 60s for the CRs and then `|| true`.  But this same PR now defers CR deletion while any PMJ in the namespace is non-terminal, which can be minutes (5-minute Restoring ceiling, ~10-minute overall).  So with a migration in flight: 60s pass, kubectl exits non-zero, `|| true` eats it, we delete the controller with the CR still Terminating, and CRD deletion hangs exactly as before - now silently and after a minute's wait.  I assume the `|| true` is for the CRD-not-installed case (`--ignore-not-found` doesn't cover a missing resource *type*).  Split the two: gate on `kubectl get crd podmigrations.podmigration.gke.io`, then delete with a timeout that outlasts a migration (600s or so) and no `|| true`, so a stuck teardown fails loud before we remove the only thing that can unstick it.

Nice to have, not blocking: the deferral is visible only in controller logs.  A Ready=False condition on the PodMigration with the in-flight count would let `kubectl describe` answer "why won't this delete".

Fix 1 and I approve.

kyber775:
Addressed Round 2 review feedback in commit `4e2cd5c`:

1. **`make undeploy` fail-loud on timeout:**
   Split the CR deletion and CRD check: `controller/Makefile`'s `undeploy` recipe now gates on `if kubectl get crd podmigrations.podmigration.gke.io >/dev/null 2>&1`, extends the timeout to 600s (outlasting the 5-minute Restoring timeout / 10-minute overall window), and removes `|| true`. If in-flight migrations or stuck finalizers prevent deletion within 600s, `make undeploy` aborts immediately with a non-zero exit code *before* removing the controller deployment or uninstalling CRDs, preventing a wedged state.
2. **`Ready=False` condition on in-flight deferral:**
   When deletion is deferred due to active migrations in flight, `podmigration_controller.go` now updates the `Ready` status condition to `Status: False`, `Reason: DeletionBlockedByInFlightMigrations`, and `Message: Deletion deferred: N migration(s) in flight (<names>)`. `kubectl describe podmigration <name>` now clearly explains why deletion is waiting.
3. **Unit test assertion:**
   Updated `TestPodMigrationReconciler_Finalizer_PostponesDeleteWhenMigrationsInFlight` to assert the `Ready=False` condition and its reason while migrations are active.

bnaylor (approved):
Round 3, at 4e2cd5c.  The undeploy recipe does the right thing now: I ran it against a fake kubectl that times out the CR delete, and make stopped with the error visible and never got to `delete -f deploy.yaml`.  600s, no `|| true`, CRD-gated.  The Ready=False condition with the job names is a nice touch for `kubectl describe`.  Manifests and tests are clean.

One nit, not worth another round: the deferral status write isn't gated on `SetStatusCondition`'s changed bool, so it writes every 5s tick while deferred.  Same pattern as the InvalidStorageLocation condition in that file, so fix both or neither whenever someone's in there.

Approving.

ahahadelyaly:
# Review comment — PR #35



---

Thanks for this — the finalizer approach is the right shape, and I want to call
out two things you got right that are easy to get wrong: the cleanup ordering is
**delete-dependents → then remove-finalizer** (so a crash mid-deletion leaks
nothing), and the finalizer is added *before* the PSSC is created (so there's no
orphan window). I checked both specifically and they're correct.

Two blocking issues before this can merge, plus a rebase.

## Blocking

### 1. Deletion deferral has no upper bound — this can wedge a namespace permanently

In `podmigration_controller.go`:

```go
if len(inFlight) > 0 {
    meta.SetStatusCondition(...)
    if err := r.Status().Update(ctx, config); err != nil { ... }
    return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
}
```

There's no ceiling on this. If any PMJ in the namespace gets stuck in a
non-terminal phase — snapshot never completes, node goes away mid-migration,
controller loses the watch — the finalizer never clears, the `PodMigration` sits
in `Terminating` forever, and **namespace deletion blocks behind it**. Recovering
requires a manual `kubectl patch` to strip the finalizer, which is exactly the
situation finalizers are supposed to save operators from.

The commit message says "fail loud on undeploy timeout", but the only timeout in
this PR is `--timeout=600s` in the Makefile shell command. The controller itself
has no timeout at all, so it can't distinguish "genuinely hung" from "slow but
progressing."

Suggested fix — bound it on the deletion timestamp:

```go
const maxDeletionDeferral = 10 * time.Minute // matches the durable snapshot timeout

if len(inFlight) > 0 {
    if time.Since(config.DeletionTimestamp.Time) > maxDeletionDeferral {
        logger.Error(nil, "Deletion deferral exceeded maximum; forcing storage cleanup",
            "name", config.Name, "inFlight", inFlight)
        r.Recorder.Eventf(config, corev1.EventTypeWarning, "DeletionDeferralTimeout",
            "Forcing cleanup after %s with %d migration(s) still in flight",
            maxDeletionDeferral, len(inFlight))
        // fall through to cleanup
    } else {
        // ... existing defer path
    }
}
```

Whatever ceiling you pick, the important part is that there *is* one and that it
emits a Warning event so it's visible in `kubectl describe`.

### 2. The in-flight check blocks on unrelated migrations

```go
r.List(ctx, &pmjList, client.InNamespace(req.Namespace))
```

This counts **every** PMJ in the namespace, not just the ones belonging to this
`PodMigration`. In any namespace running more than one migrating workload,
deleting `PodMigration` A will block on workload B's in-flight migration, which
has nothing to do with A's storage config.

This also meaningfully widens the blast radius of issue #1 above — the more PMJs
in the namespace, the more likely something is stuck, and the more likely you
wedge.

Please filter to PMJs owned by this `PodMigration` (owner reference or label
selector) before computing `inFlight`.

## Non-blocking, but please address

### 3. Status write amplification in the deferral loop

`meta.SetStatusCondition` returns a bool telling you whether anything actually
changed, and it's being discarded — so `r.Status().Update()` fires
unconditionally every 5 seconds for the entire duration of the deferral, even
when the condition is byte-identical.

```go
changed := meta.SetStatusCondition(&config.Status.Conditions, metav1.Condition{
    Type:               "Ready",
    Status:             metav1.ConditionFalse,
    Reason:             "DeletionBlockedByInFlightMigrations",
    Message:            fmt.Sprintf("Deletion deferred: %d migration(s) in flight (%s)",
                            len(inFlight), strings.Join(inFlight, ", ")),
    ObservedGeneration: config.Generation,
})

if changed {
    if err := r.Status().Update(ctx, config); err != nil {
        return ctrl.Result{}, err
    }
}
return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```

Worth noting the `Message` embeds a joined list of PMJ names, so as jobs complete
the string churns and the condition genuinely does change — which makes gating on
`changed` necessary but not sufficient. Consider dropping the name list from the
message (keep the count) and logging the names instead.

### 4. Conflict errors are swallowed, defeating backoff

Same block: the status update error is logged and then the function returns
`nil` with a fixed 5s requeue. That discards controller-runtime's exponential
backoff and turns a conflict storm into a fixed-rate retry storm. Return the
error and let the workqueue handle it.

### 5. PSSC is deleted by name without asserting ownership

`deleteStorageResources` constructs the name and issues `r.Delete` with no
read-back. The `sha256(namespace/name)` scheme does correctly prevent
cross-namespace collisions — I verified that, it's not the concern here.

The narrow case is delete-then-recreate of the *same* `namespace/name`: a new
`PodMigration` creates a PSSC at the same hashed name while the old object's
finalizer is still draining, and the old finalizer then deletes the new object's
PSSC.

Cheap fix: stamp `podmigration.gke.io/owner-uid: <uid>` on the PSSC at creation,
then `Get` and compare before deleting.

### 6. `make undeploy` aborts before it finishes tearing down

```makefile
kubectl delete podmigrations.podmigration.gke.io --all --all-namespaces \
    --ignore-not-found=true --wait=true --timeout=600s
```

No `|| true`, so on timeout `make` exits non-zero **before** reaching
`kubectl delete -f deploy.yaml` and `make uninstall`. The operator is left with a
running controller, wedged CRDs, and a half-removed deployment.

This composes badly with issue #1: the unbounded deferral is what makes the 600s
expiry likely in the first place, and this line is what makes the expiry
destructive. Either allow the failure (`|| true`) and continue teardown, or drop
`--wait=true` and let the finalizers drain asynchronously.

### 7. Downgrade path needs a documented escape hatch

A cluster rolled back to a pre-#35 controller image gets a binary that doesn't
know the `podmigration.gke.io/storage-cleanup` finalizer string and will never
remove it — every `PodMigration` deleted while rolled back wedges in
`Terminating`.

You can't patch old binaries, so this needs to be a documented operator step in
`README.md`:

```bash
kubectl patch podmigration <name> -n <ns> \
  --type=json -p='[{"op":"remove","path":"/metadata/finalizers"}]'
```

### 8. Please rebase onto current `main`

This branch is based on the semantic-versioning commit and predates the
multi-replica HA / leader election work that has since landed, plus one commit
after that.

It merges cleanly and I confirmed the merge result does **not** regress any of
the HA settings — `replicas: 2`, `--leader-elect=true`, the PDB, and the
`kube-system` webhook exclusions all survive intact, and your new
`podmigrations/finalizers` RBAC rule lands alongside them. So there's no
correctness problem with the merge itself.

The concern is test coverage: none of the 428 new test lines have ever executed
against a 2-replica leader-elected controller. The finalizer path specifically
interacts with leader failover mid-deletion, and I'd like to see that exercised
on the current base before this goes in.

## Test gaps

The new tests cover the happy paths and the `NotFound` PSSC case well. Missing:

- **Deferral timeout expiry** — assert cleanup is forced once the ceiling is
  exceeded, rather than deferring forever.
- **Conflict during finalizer removal** — use an intercepting fake client that
  returns a 409 on the first `Update`, assert the reconciler recovers.
- **Owner-scoped in-flight filtering** — a second, unrelated `PodMigration` with
  an in-flight PMJ in the same namespace must *not* block deletion of the first.

## Suggested merge checklist

- [ ] Bound the deletion deferral and emit a Warning event on expiry
- [ ] Scope the in-flight PMJ list to this `PodMigration`
- [ ] Gate the status write on `changed`; return conflict errors
- [ ] Add `owner-uid` label + get-and-compare before PSSC delete
- [ ] Make `undeploy` resilient to timeout
- [ ] Document the downgrade `kubectl patch` in `README.md`
- [ ] Rebase onto current `main` and re-run the full suite
- [ ] Add the three test cases above

kyber775:
Addressed Round 3 review feedback in commit `4841ddf` (rebased onto `upstream/main`):

1. **Bounded deletion deferral with warning event:**
   Added `maxDeletionDeferral = 10 * time.Minute`. If in-flight migrations remain after 10 minutes, `podmigration_controller.go` logs an error, records a `DeletionDeferralTimeout` Warning event via `r.Recorder.Eventf`, and falls through to force storage resource cleanup so namespaces cannot wedge indefinitely.
2. **Owner-scoped in-flight PMJ filtering:**
   Added `isPMJOwnedByPodMigration` helper to filter in-flight PMJs by `OwnerReferences` (matching `PodMigration` name/UID) or `pod-migration.gke.io/podmigration-name` / `pod-migration.gke.io/owner-uid` labels. In-flight migrations belonging to other workloads in the same namespace no longer block deletion.
3. **Status write amplification & backoff:**
   Gated `r.Status().Update` on `changed := meta.SetStatusCondition(...)`. Simplified the condition message to `"Deletion deferred: N migration(s) in flight"` (logging individual job names in controller logs to eliminate message churn). Returned status update errors directly so controller-runtime workqueue backoff handles API conflicts rather than swallowing them.
4. **PSSC ownership verification before deletion:**
   Stamped `podmigration.gke.io/owner-uid: <uid>` label on `PodSnapshotStorageConfig` during creation/sync. In `deleteStorageResources`, the reconciler gets the PSSC and verifies that the owner UID matches before issuing the delete call, preventing delete-then-recreate races from deleting a newer instance's storage configuration.
5. **Resilient `make undeploy`:**
   Added `|| true` to the `kubectl delete podmigrations` recipe in `controller/Makefile` so teardown continues to delete `deploy.yaml` and uninstall CRDs even if the 600s wait expires.
6. **Documented rollback escape hatch:**
   Added operator guidance to `README.md` detailing the manual JSON patch command when rolling back to an older controller image lacking finalizer support.
7. **Rebased onto `upstream/main`:**
   Cleanly rebased onto `upstream/main` (`5bf2945`) with all 12 controller test suites passing 100%.
8. **Regression tests added:**
   - `TestPodMigrationReconciler_Finalizer_DeferralTimeoutForcesCleanup`: asserts forced cleanup and `DeletionDeferralTimeout` warning event emission upon ceiling expiry.
   - `TestPodMigrationReconciler_Finalizer_UnrelatedMigrationsDoNotBlock`: asserts active jobs from another PodMigration do not block deletion.
   - `TestPodMigrationReconciler_Finalizer_ConflictRetryDuringFinalizerRemoval`: asserts recovery from 409 conflict during finalizer removal.
   - `TestPodMigrationReconciler_Finalizer_RecreatedPSSCNotDeleted`: asserts PSSC with mismatched owner UID is preserved.

ahahadelyaly:
# Review comment — PR #35 (Round 3)



---

Re-reviewed Round 3 against the merge base. All six items are genuinely fixed —
I traced the control flow rather than just checking the symbols are present, and
the logic is correct. Approving.

## Verification

**The deferral timeout actually falls through.** This was my main worry: a
timeout that logs, emits an event, and then still returns a requeue would look
like a fix while wedging exactly as before. It doesn't — the timeout branch sets
no `return`, so control exits the `len(inFlight) > 0` block and lands on
`deleteStorageResources` + `RemoveFinalizer`. Correct.

Also confirmed on that path:

- `!config.DeletionTimestamp.IsZero()` guards the comparison, so no nil deref.
- `if r.Recorder != nil` guards the event emission, and `cmd/main.go` wires a
  real recorder.
- `changed := meta.SetStatusCondition(...)` now gates the write.
- The status-update error is **returned**, so the workqueue's exponential backoff
  applies instead of a fixed 5s retry.
- `isPMJOwnedByPodMigration(&pmj, config)` scopes the in-flight check, so an
  unrelated workload's migration no longer blocks this CR's deletion.
- `deleteStorageResources(..., config.UID)` threads the UID through for the
  ownership assertion before deleting the PSSC.

**One thing you fixed that nobody asked for**, and it's the right call: the
condition `Message` no longer embeds the joined list of PMJ names, just the
count. Gating on `changed` alone wouldn't have been sufficient while that string
churned as jobs completed — each completion would have flipped the condition and
triggered another write. Dropping the list closes that properly. The names are
still in the structured log where they belong.

## Two nits, neither blocking

**1. The nil-`Recorder` guard is untested.** Every test wires a
`record.NewFakeRecorder(...)`, so nothing currently fails if a future refactor
strips the `if r.Recorder != nil` check. A test constructing the reconciler with
`Recorder: nil` and asserting the timeout path still completes the cleanup would
pin the invariant. Cheap insurance on a path that only executes in the failure
case, which is where a panic hurts most.

**2. The in-flight `List` is namespace-wide with in-memory filtering.** Correct
as written, but it's an O(n) list on every deletion reconcile. Fine at current
scale; worth a field index or label selector if per-namespace PMJ counts grow.
Not something I'd hold the PR for.

## Merge sequencing

This is rebased onto current `main` and needs no further changes, so I'd merge it
first. The other in-flight PR touches a disjoint set of files (the PMJ and pod
gate controllers, versus this PR's `PodMigration` controller), so there's no
textual conflict either way — but it still needs changes, and merging this one
first means it only has to rebase once onto a known-good base.

LGTM.

ahahadelyaly (approved):

PR #36, opened 2026-09-16
fix(controller): recover stranded PMJ when recorded consumer pod is deleted (Closes #25)

### LINKED ISSUE #25: [P2] Stranded PMJ Recovery: verify recorded consumer pod exists before honoring Consumed
### Problem
In `PodGateReconciler`, marking `job.Status.Consumed = true` occurs before the Pod update that removes the scheduling gate.
If a replacement pod is deleted in the millisecond window between these two writes (e.g. node crash, preemptive deletion), the PMJ remains permanently stranded as `Consumed: true` with a dead `RestoredPodUID`. Subsequent candidate pods skip the PMJ via the cold-start bypass, stranding an otherwise valid snapshot.

### Proposed Solution
In `PodGateReconciler`, before honoring `Consumed: true`:
1. Check if the pod with UID `job.Status.RestoredPodUID` still exists in the API server / cache.
2. If the recorded consumer pod no longer exists (died before restoring), recover the PMJ by resetting `job.Status.Consumed = false` and `job.Status.RestoredPodUID = ""` so subsequent candidate pods can adopt and restore it.

### References
- Identified during PR #11 review: https://github.com/gke-labs/pod-migration/pull/11#issuecomment-5416618113

kyber775:
## Summary

Recovers stranded `PodMigrationJob` (PMJ) resources in `PodGateReconciler` when a recorded consumer pod was deleted in the race window before scheduling gate release, enabling subsequent candidate replacement pods to adopt and restore the durable snapshot. Scoped strictly via `GateReleased` to prevent snapshot replay/state rollback.

## Why This Change

In `PodGateReconciler`, marking `job.Status.Consumed = true` and recording `RestoredPodUID` occurs before updating the candidate pod to remove the scheduling gate.
If the replacement pod is deleted in the window before its scheduling gate is released (e.g. node crash, preemptive deletion, or pod eviction), the PMJ remained permanently stranded as `Consumed: true` with a dead `RestoredPodUID`.
When subsequent candidate pods arrived, they observed `job.Status.Consumed && job.Status.RestoredPodUID != string(pod.UID)` and released with the cold-start bypass (`podsnapshot.gke.io/ps-name = ""`), stranding an otherwise valid snapshot.
Furthermore, if the consumer pod is still terminating (e.g. during eviction with grace period), releasing the candidate pod immediately with cold-start bypass permanently loses the snapshot; requeuing keeps the candidate gated until the terminating consumer finishes and disappears from the cluster.

## What Changed

- **`controller/api/v1alpha1/podmigrationjob_types.go`**:
  - Added `GateReleased bool` to `PodMigrationJobStatus`. Written after the candidate pod's scheduling gate removal update succeeds on the API server.
- **`controller/config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml`**:
  - Updated CRD schema manifests with `GateReleased` status field definition.
- **`controller/internal/controller/pod_gate_controller.go`**:
  - Added `GateReleased` check to ensure recovery only occurs if `!job.Status.GateReleased` (i.e. the recorded consumer never had its gate released). If the gate was already released, subsequent candidate pods receive the cold-start bypass to prevent state rollback to checkpoint time.
  - Updated recovery write sequence: writes PMJ status first (setting `Consumed = true`, `RestoredPodUID = pod.UID`, `RestoredPodName = pod.Name`, resetting `RestoringStartTime = &now`, and `GateReleased = false`) before clearing `util.AnnotationMismatchSince` on metadata, preventing status subresource reversions.
  - Added requeue (`ctrl.Result{RequeueAfter: 5 * time.Second}`) when the recorded consumer pod still exists (e.g. terminating), holding the candidate in its scheduling gate until the prior consumer pod finishes terminating.
  - Optimized `checkConsumerPodExists` fallback when `RestoredPodName == ""` to query `PodAssignedPMJIndex` in O(1) instead of listing the whole namespace.
  - Verified cache hits against `apiReader` to guard against stale cache hits where a pod was deleted on the API server.
  - Added kubebuilder RBAC marker for `podmigrationjobs/status` (`get;update;patch`).
- **`controller/internal/controller/pod_gate_controller_test.go`**:
  - `TestPodGateReconciler_Reconcile_RecordsRestoredPodNameAndUID`: verifies `GateReleased == true` after un-gating, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_RecoversAndAdopts`: sets prior `RestoringStartTime` and verifies it is reset on adoption, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_TerminatingConsumerDoesNotRecover`: asserts candidate pod remains gated and reconciler requeues with 5s.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_CacheLagConsumerStillOnAPIServer`: asserts candidate remains gated and requeues with 5s.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_ConfirmedDeletedOnAPIServer`: asserts candidate un-gates and adopts once deleted on API server, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_ClearsMismatchSinceAnnotation`: asserts annotation cleared and status fields updated, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_SucceededPMJ_NeverReRestored`: asserts completed PMJ never re-restores, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_UIDMismatchRecovers`: asserts recovery on recreated pod instance with new UID, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_EmptyRestoredPodNameRecovers`: asserts O(1) index recovery when `RestoredPodName` is empty, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_EmptyRestoredPodName_ConsumerStillExists`: asserts candidate remains gated and requeues 5s when consumer still exists matching index.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_GateAlreadyReleasedDoesNotRecover`: asserts consumer pod that died after un-gating cannot be re-restored, asserts `reconcile.Result`.
  - `TestPodGateReconciler_Reconcile_StrandedPMJ_StaleCacheHitConfirmedDeletedOnAPIServer`: asserts recovery succeeds despite stale cache hit when API server confirms NotFound, asserts `reconcile.Result`.

## Context

Closes #25
Follow-up to PR #11 code review finding 2.

## Testing

- Sabotage verification: verified that `TestPodGateReconciler_Reconcile_StrandedPMJ_EmptyRestoredPodName_ConsumerStillExists` fails on sabotage.
- `cd controller && go test -v -count=1 ./...` (all unit test suites pass).
- `cd controller && make verify-generate` (manifests, deepcopy, deploy assembly up to date).

## Live validation

Not live-tested on GKE cluster: this is pure controller reconcile logic covering premature pod deletion before un-gating, fully verified via controller-runtime unit tests and envtest integration tests against simulated API server responses.

## Self-Review

Isolated pre-PR review conducted via subagent `deleg_aa2b1c36` and Round 1 review by @bnaylor:
- **Finding 1 ([CRITICAL])**: In-memory status update was reverted by subsequent metadata `r.Update(ctx, job)` clearing `AnnotationMismatchSince`, because status subresource PUT decodes stored server status back into `job`.
  - *Disposition*: Fixed; write status update first (assigning candidate pod and resetting `RestoringStartTime`), then clear the annotation on metadata.
- **Finding 2 ([CRITICAL])**: "Consumer exists" released the candidate pod immediately with cold-start bypass, losing the snapshot on terminating consumers.
  - *Disposition*: Fixed; requeue candidate with `5s` backstop while consumer exists, holding candidate in its gate until terminating consumer disappears.
- **Finding 3 ([CRITICAL])**: Recovering any `Restoring` PMJ risked re-restoring a snapshot whose consumer pod already un-gated and ran, causing application state rollback to checkpoint time.
  - *Disposition*: Fixed; added `GateReleased` status field written after pod update un-gating succeeds. Recovery is strictly gated to `!job.Status.GateReleased`.
- **Finding 4 ([WARNING])**: `RestoringStartTime` was not reset on adoption, risking premature timeout if consumer died near the 5-minute ceiling.
  - *Disposition*: Fixed; reset `RestoringStartTime = &now` on adoption.
- **Finding 5 ([WARNING])**: `RestoredPodName == ""` fallback scanned the entire namespace.
  - *Disposition*: Fixed; optimized to O(1) query using `PodAssignedPMJIndex`.
- **Finding 6 ([SUGGESTION])**: Unit tests discarded `reconcile.Result`.
  - *Disposition*: Fixed; asserted `reconcile.Result` across all test cases.

## Risk & Rollout

- **Blast radius**: Scoped strictly to gated replacement pods encountering an assigned PMJ in `Restoring` phase whose recorded consumer pod has already been deleted from the API server before gate release.
- **Rollback**: Standard controller deployment rollback; does not introduce backward-incompatible CRD mutations.


bnaylor:
Round 1, at 6bd00e0.  This is the right shape for #25, and the cache-vs-APIReader handling on the miss side is careful.  The sabotage check in the description is the kind of thing I want to see more of.  Four things, the first one reproduced:

1. **The annotation Update reverts the status reset.**  [Lines 209-214](https://github.com/gke-labs/pod-migration/blob/6bd00e0b3ff1eae2b4d8d1ca515822fcf606aa3a/controller/internal/controller/pod_gate_controller.go#L207-L219): you zero Consumed/RestoredPodUID/RestoredPodName in memory, then `r.Update(ctx, job)` to drop the mismatch-since annotation.  PMJ has a status subresource, so the server ignores status on that PUT and hands back the stored status, and the client decodes it into `job`.  Consumed is true again with the dead UID, so the `if !job.Status.Consumed` block at line 230 never runs.  The candidate gets un-gated with the real ps-name, but the PMJ still points at the dead pod: the PMJ reconciler waits for a pod that doesn't exist until the 5-minute ceiling and marks SucceededWithoutRestore, and every later candidate sees "consumer gone" and adopts again.  Reproduced by adding a RestoredPodUID assertion to [ClearsMismatchSinceAnnotation](https://github.com/gke-labs/pod-migration/blob/6bd00e0b3ff1eae2b4d8d1ca515822fcf606aa3a/controller/internal/controller/pod_gate_controller_test.go#L1550-L1552) - it fails with `uid-dead`.  That test asserts only the annotation, which is how this got through.  Fix: do the status write (with the new consumer) first, then clear the annotation, and make that test check the status fields.

2. **"Consumer exists" releases the candidate irreversibly, with no requeue.**  [Lines 194-205](https://github.com/gke-labs/pod-migration/blob/6bd00e0b3ff1eae2b4d8d1ca515822fcf606aa3a/controller/internal/controller/pod_gate_controller.go#L193-L205).  Terminating counts as existing (by design, per the test), and a cache hit with the right UID is never re-confirmed live.  Both are transient states.  Deployment case: consumer A is being evicted with a 30s grace, the RS creates candidate B right away, PodGate sees A still present, B gets the cold-start bypass forever, A finishes terminating, the PMJ waits out the ceiling.  The snapshot is lost on the most common path.  B is gated, so waiting costs nothing: requeue (5s or so) while the consumer is present, recover once it's gone.  That also prevents the split-brain the Terminating rule was for.

3. **Phase==Restoring isn't "died before un-gating".**  The comment and #25 both scope recovery to the write-order window, but Restoring lasts until the pod is Ready, the ceiling, or a 30s mismatch - up to 5 minutes in which the consumer may have been un-gated, restored and running.  Recovering then restores the same snapshot into a second pod, ie a rollback to checkpoint time, which is the class the Consumed guard in #11 exists to stop.  Two ways out: record that the gate was released (a status field written after the pod Update succeeds) and recover only when it's absent - that's exactly the #25 window; or decide that re-restoring any unverified consumer is the policy and say so in the comment and in the Consumed doc in the types file (which still says "once true").  I lean to the marker.  Either way, decide it out loud.

4. **RestoringStartTime isn't reset on adoption.**  Consumer dies at minute 4, candidate adopted, ceiling fires at minute 5 with Consumed=true and no gated-claimant deferral -> SucceededWithoutRestore while the candidate is mid-restore.  Reset it when you reassign the consumer.  None of the new tests set it.

Smaller:

- All the new tests discard the reconcile Result.  Same note as the #12 post-merge review.
- The RestoredPodName=="" fallback lists the whole namespace.  The assigned-pmj index from #33 gives the same answer in O(1), since the consumer carries the annotation.
- Only two of the new tests wire a distinct APIReader; the rest fall back to the cache, which is fine for what they test, but the stale-hit direction has no test at all.

1, 2 and 4, plus a decision on 3, and I approve.

kyber775:
Addressed Round 1 review feedback in commit `9e78912`:

1. **Status reversion fix:** Re-ordered the writes so `r.Status().Update(ctx, job)` (assigning the new candidate consumer pod and resetting `RestoringStartTime`) executes **before** `r.Update(ctx, job)` clears `util.AnnotationMismatchSince`. Updated `TestPodGateReconciler_Reconcile_StrandedPMJ_ClearsMismatchSinceAnnotation` to assert that `job.Status.RestoredPodUID == string(candidatePod.UID)` and `GateReleased == true`.
2. **Requeue while consumer exists:** When `consumerExists == true` (e.g. previous consumer is in grace period / terminating), the reconciler now logs and returns `ctrl.Result{RequeueAfter: 5 * time.Second}` without releasing the candidate pod. The candidate stays safely held in Pending behind its scheduling gate until the terminating pod finishes and disappears from the API server, at which point it adopts the PMJ snapshot. Updated `TestPodGateReconciler_Reconcile_StrandedPMJ_TerminatingConsumerDoesNotRecover` to assert that the gate remains on the candidate pod.
3. **Decision on recovery scope (Item 3 — Option A: `GateReleased` marker):**
   - Added `GateReleased bool` to `PodMigrationJobStatus` (regenerated CRD schema and manifests).
   - Marked `GateReleased = true` immediately after the Pod update removing the scheduling gate succeeds.
   - Stranded PMJ recovery is strictly gated on `!job.Status.GateReleased`. If `GateReleased == true`, the previous consumer was already un-gated and executing; subsequent candidate pods release with cold-start bypass, preserving PR #11 single-use isolation and preventing application state rollback to checkpoint time.
   - Added unit test `TestPodGateReconciler_Reconcile_StrandedPMJ_GateAlreadyReleasedDoesNotRecover`.
4. **`RestoringStartTime` reset:** Stamped `RestoringStartTime = &metav1.Time{Time: time.Now()}` upon adoption so the new candidate gets the full 5-minute restoration timeout window instead of inheriting the dead consumer's elapsed clock.
5. **Stale cache-hit verification:** In `checkConsumerPodExists`, matching-UID cache hits now re-confirm live against `r.apiReader()` to guard against stale informer caches. Added unit test `TestPodGateReconciler_Reconcile_StrandedPMJ_StaleCacheHitConfirmedDeletedOnAPIServer`.

kyber775:
Addressed Round 1 review feedback in commit `4f357d3` (and synced with `upstream/main`):

1. **Status write ordering vs annotation update:**
   Reversed the operation order: we now execute `r.Status().Update(ctx, job)` first (assigning the candidate pod as consumer, resetting `RestoringStartTime`, and initializing `GateReleased: false`), and only then clear `util.AnnotationMismatchSince` via `r.Update(ctx, job)`. Updated `TestPodGateReconciler_Reconcile_StrandedPMJ_ClearsMismatchSinceAnnotation` to explicitly assert that `updatedPMJ.Status.RestoredPodUID == "uid-candidate"` and `GateReleased == true`.

2. **Requeue on terminating consumer instead of cold-start bypass:**
   Updated `pod_gate_controller.go`: when `consumerExists` is true, the reconciler returns `ctrl.Result{RequeueAfter: 5 * time.Second}`. The candidate pod remains safely held in `Pending` by its scheduling gate until the terminating pod finishes and disappears from the API server, at which point the next reconcile adopts the PMJ. Tested in `TestPodGateReconciler_Reconcile_StrandedPMJ_TerminatingConsumerDoesNotRecover`.

3. **Phase==Restoring vs died-before-un-gating (Marker Decision):**
   Adopted the explicit marker approach: added `GateReleased bool` to `PodMigrationJobStatus` (and regenerated CRD schemas). It is stamped `true` in a status update immediately after the candidate pod's un-gating `Update` succeeds. Recovery is strictly gated to `job.Status.Phase == Restoring && !job.Status.GateReleased`. If `GateReleased` is already true, subsequent candidate pods receive the cold-start bypass, preserving single-use isolation and preventing state rollback to checkpoint time. Added `TestPodGateReconciler_Reconcile_StrandedPMJ_GateAlreadyReleasedDoesNotRecover`.

4. **RestoringStartTime reset on adoption:**
   Reset `job.Status.RestoringStartTime = &now` on adoption in `pod_gate_controller.go`. Updated `TestPodGateReconciler_Reconcile_StrandedPMJ_RecoversAndAdopts` to initialize a 4-minute-old timestamp and verify that `RestoringStartTime` is reset upon adoption.

**Smaller items addressed:**
- **Reconcile Result assertions:** Updated all test cases to assert `reconcile.Result` (verifying empty result on un-gate and `RequeueAfter: 5s` on waiting).
- **O(1) index on empty `RestoredPodName`:** Replaced the namespace scan with `client.MatchingFields{PodAssignedPMJIndex: job.Name}` in `checkConsumerPodExists`, paired with an `apiReader` check. Added `TestPodGateReconciler_Reconcile_StrandedPMJ_EmptyRestoredPodName_ConsumerStillExists` with sabotage verification.
- **Stale cache hit test:** Verified with `TestPodGateReconciler_Reconcile_StrandedPMJ_StaleCacheHitConfirmedDeletedOnAPIServer`.
- **Rebase onto `upstream/main`:** Rebased onto `upstream/main` (`b3677f3`), with the full unit test suite passing 100%.

bnaylor:
@ahahadelyaly similar question for the above change.  Makes sense to me but is a policy decision.

bnaylor:
Round 2, at a53d05b.  Everything from round 1 landed, including the marker route on item 3, and I'm fine with that decision - I've asked ahaha for her take as well.  Verified: build/vet/test clean, `verify-generate` clean for the new CRD field, the ClearsMismatchSince test now asserts the consumer UID, the stale-hit direction has a test, and the index constants are the same string (I checked because the fallback uses a different name than ResolveCollision).  Re-anchoring RestoringStartTime at consumption for the normal path too is a change to #29's semantics, but a good one: the ceiling now measures restore time rather than queue time.  Say that in the commit message.

Two things, one blocking:

1. **The GateReleased write is fire-and-forget, and nothing ever retries it.**  [Lines 266-273](https://github.com/gke-labs/pod-migration/blob/a53d05b0ad31219081590d5f4322a8ac9e5f0ce7/controller/internal/controller/pod_gate_controller.go#L264-L275): after the pod Update removes the gate, the status write that sets GateReleased=true logs on error and returns success.  The next reconcile of that pod exits at the [missing-gate early return](https://github.com/gke-labs/pod-migration/blob/a53d05b0ad31219081590d5f4322a8ac9e5f0ce7/controller/internal/controller/pod_gate_controller.go#L66-L76) before any status logic, and no other code sets the field.  One Conflict on that write - and the PMJ reconciler is writing status and annotations on the same object from 50 workers during Restoring - leaves GateReleased false forever, and the marker that exists to stop re-restoring a live consumer's snapshot is now off for that PMJ.  Reproduced with an interceptor that fails only that write: second reconcile, still false.  The robust fix is to make it self-healing rather than retried: the PMJ reconciler already Gets the consumer pod by name every 2s in Restoring, so have it stamp GateReleased when it sees that pod (name+UID match) without the gate.  Then a missed write on the PodGate side costs a couple of seconds instead of the invariant.

2. **Same shape, smaller: the mismatch-since clear isn't retried either.**  If the annotation `r.Update` at [lines 228-234](https://github.com/gke-labs/pod-migration/blob/a53d05b0ad31219081590d5f4322a8ac9e5f0ce7/controller/internal/controller/pod_gate_controller.go#L227-L235) conflicts, the reconcile retries, but by then RestoredPodUID is already this pod's, so the recovery block (and the clear) is skipped.  The stale timestamp then makes the next genuine mismatch on the adopted pod fast-fail instantly instead of after 30s.  Clearing it in the PMJ reconciler when name and UID match again would fix both this and the ownership oddity of PodGate editing PMJ metadata.

Housekeeping: the branch has each commit twice plus a merge of itself.  Squash-merge hides it, but please don't do that again.  And the Consumed doc comment in the types file still says "once true"; it isn't anymore.

Fix 1 (2 if you're in there anyway) and I approve.

kyber775:
Addressed Round 2 review feedback in commit `3c868c7`:

1. **Self-healing `GateReleased` in PMJ reconciler:**
   In `podmigrationjob_controller.go`, when the replacement pod is evaluated in `Restoring` phase (name+UID match) and `podHasMigrationGate(replacementPod) == false`, the PMJ reconciler stamps `job.Status.GateReleased = true`. If the fire-and-forget status update in `PodGateReconciler` encounters an API Conflict under load, the PMJ reconciler self-heals the field on its 2s polling tick, guaranteeing the marker invariant without relying on fragile uncoordinated retries.
2. **Self-healing `AnnotationMismatchSince` cleanup:**
   Transferred ownership of clearing `util.AnnotationMismatchSince` from `PodGateReconciler` to `PodMigrationJobReconciler`. When a replacement pod matches `job.Status.RestoredPodUID`, the PMJ reconciler deletes `AnnotationMismatchSince` from `job.Annotations` and updates the PMJ metadata. This eliminates cross-controller metadata mutations by `PodGateReconciler` and ensures conflicts during recovery cannot leave a stale timestamp behind.
3. **`Consumed` & `GateReleased` doc comments:**
   Updated `Consumed` and `GateReleased` field documentation in `controller/api/v1alpha1/podmigrationjob_types.go` to explicitly document the stranded recovery semantics (no longer asserting "once true") and regenerated CRD manifests via `make manifests` (verified clean with `make verify-generate`).
4. **Semantics note on `RestoringStartTime`:**
   Re-anchoring `RestoringStartTime` at consumption for the normal path as well as the recovery path updates #29's semantics so that the 5-minute safety ceiling measures active restore time rather than queue time.
5. **Unit test additions:**
   - Added `TestPodMigrationJobReconciler_Restoring_SelfHealsGateReleasedWhenPodUngated`.
   - Added `TestPodMigrationJobReconciler_Restoring_GateReleasedRemainsFalseWhileGatePresent`.
   - Added `TestPodMigrationJobReconciler_Restoring_ClearsMismatchSinceWhenUIDMatches`.
   - Updated `TestPodGateReconciler_Reconcile_StrandedPMJ_AdoptsNewCandidateUIDAndReleasesGate` to verify adoption status and gate release without expecting metadata edits.

bnaylor (approved):
Round 3, at 3c868c7.  The self-heal is where it belongs: after the name+UID match, gated on GateReleased being false, using the shared gate helper, error returned so it retries.  I reproduced round 2's failure (PodGate's GateReleased write fails, then a PMJ reconcile) and the marker comes back true on the next tick.  The mismatch-since clear moved with it, PodGate no longer touches PMJ metadata, the doc comments and CRD match, and the RestoringStartTime note is in the commit message.  Build, vet, tests, verify-generate all clean.

The branch still has each commit twice plus the self-merge.  Squash on merge and it disappears.

Approving.

ahahadelyaly:
# Review comment — PR #36



---

This is tackling a real and nasty class of bug, and the `mismatch-since`
implementation in particular is careful work — I traced it and it's correct: the
annotation is stamped exactly once on first observation, it does *not* get
re-stamped on every reconcile (which would livelock the grace period forever),
and it self-heals by clearing when a correct UID rotates into the slot. That's
the subtle part and you got it right.

Two blocking issues though, and they both terminate in the same place:
**two pods restoring from one snapshot**.

## Blocking

### 1. The `GateReleased` status update can be silently lost

In `pod_gate_controller.go`, the pod is ungated via `r.Update`, then
`GateReleased = true` is written with a bare `r.Status().Update()` whose error is
logged and dropped as *"non-fatal since pod is already un-gated."*

It isn't non-fatal. With 50 PMJ workers writing status to the same object, a 409
here is likely, not exotic. When it happens:

1. The pod is ungated and running — actively mutating its PV.
2. The PMJ still reads `GateReleased == false`.
3. That pod later restarts, is evicted, or is deleted.
4. Stranded-recovery sees `GateReleased == false`, concludes the consumer died
   *before* it ever left the gate, and assigns the snapshot to a new candidate.
5. The new pod restores a **stale memory checkpoint over already-mutated disk
   state**.

The single-use consumption guard doesn't save you here, because as far as the
PMJ is concerned the snapshot was never consumed.

```go
if !job.Status.GateReleased {
    err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
        latest := &pmv1alpha1.PodMigrationJob{}
        if err := r.Get(ctx, client.ObjectKeyFromObject(job), latest); err != nil {
            return err
        }
        if latest.Status.GateReleased {
            return nil // already recorded by another writer
        }
        latest.Status.GateReleased = true
        return r.Status().Update(ctx, latest)
    })
    if err != nil {
        logger.Error(err, "Failed to record GateReleased after retries", "job", job.Name)
        return ctrl.Result{}, err // must not be swallowed
    }
}
```

This is a state-critical lock, not bookkeeping — it needs to be durable before
the reconcile is allowed to succeed.

### 2. `checkConsumerPodExists` trusts a cache-only read when `RestoredPodName` is empty

The `RestoredPodName != ""` path correctly falls back to a live `APIReader` read
— that's the right pattern and matches what the replacement webhook already does.

The `RestoredPodName == ""` path (reachable for PMJs persisted by earlier
revisions) does an index-backed `r.List()`, which is served **only** from the
informer cache. On a lagging cache that returns empty, the function reports
"consumer is dead" and recovery fires against a pod that is alive and running.

Same destination as issue #1: second gate released, second restore, two live
copies of a stateful workload sharing one checkpoint.

**On the fix** — I want to flag two tempting approaches that are both wrong:

- Returning "assume alive" unconditionally trades corruption for a **permanent
  strand**: the PMJ can then never be recovered, which is the bug this PR exists
  to fix.
- Returning an error to requeue until the cache settles **thrashes forever** when
  the pod is genuinely gone — same problem.

`APIReader` *can* serve `List`; what it can't do is use a field index. So do the
live list and filter in memory:

```go
// No RestoredPodName recorded — fall back to a LIVE namespace list.
// APIReader bypasses the cache, so an empty result here is authoritative.
podList := &corev1.PodList{}
if err := r.APIReader.List(ctx, podList, client.InNamespace(namespace)); err != nil {
    return false, err
}
for i := range podList.Items {
    if isAssignedConsumer(&podList.Items[i], job) { // same predicate the index encodes
        return true, nil
    }
}
return false, nil
```

That fails closed against stale caches without giving up the ability to recover.
It's a more expensive call, but it only runs on the legacy no-`RestoredPodName`
path, so the cost is bounded.

## Non-blocking, but please address

### 3. `mismatch-since` is evaluated against wall-clock across HA replicas

The 30s grace period uses `time.Since()` on an RFC3339 string stamped by
whichever replica held the lease at the time. If the lease moves to a replica
whose clock runs ahead by more than the grace period, that replica immediately
evaluates the window as expired and fast-fails the migration to
`SucceededWithoutRestore`.

Node clock skew on the order of tens of seconds is uncommon but absolutely
happens. Either add a skew-absorption buffer, or key the grace period off
reconciler requeue timing rather than comparing absolute wall-clock across
processes.

Worth also confirming the behaviour when the annotation is *absent* on CRs
written by an older controller — make sure a missing value isn't parsed into a
zero time and treated as infinitely old, which would trigger immediate spurious
recovery on the first reconcile after upgrade.

### 4. Test gaps

~1,300 new test lines is a lot of coverage and the happy paths are solid, but
the two failure domains this PR actually introduces aren't exercised:

- `TestPodGate_GateReleased_ConflictRetry` — use an intercepting fake client
  (`interceptor.Funcs`) that returns a 409 on the first `Status().Update`, assert
  the reconciler retries and eventually persists `GateReleased`. The absence of
  this test is why issue #1 got this far.
- `TestPodGate_EmptyRestoredPodName_UsesLiveReader` — assert that with an empty
  informer cache and a pod present only on the live API, recovery does **not**
  fire.

### 5. Commit history

The branch carries a merge commit plus three duplicated commit pairs (the
"recover stranded PMJ", "Round 1 feedback", and "adversarial review findings"
commits each appear twice).

I diffed the net tree against the individual commits and the **result is
coherent** — nothing was double-applied and no earlier review fix was dropped, so
this isn't a correctness problem. But it makes `git blame` on this file
significantly harder to read. Please `git rebase -i` down to atomic commits and
force-push before merge.

## Verified — no action needed

- **API compatibility.** The added field is an optional scalar, so existing
  persisted CRs stay valid and no `zz_generated.deepcopy.go` regeneration is
  required (scalars are covered by the generated struct assignment). CI's
  generate-freshness gate will confirm.
- **`mismatch-since` semantics.** Set once, not re-stamped, cleared correctly on
  UID rotation within the grace window. No livelock.

## One cross-PR note

PR #32 changes origin-pod termination to use the `eviction` subresource, which
changes *when* the origin pod actually disappears. That's precisely the signal
this PR's "consumer pod no longer exists" predicate keys off. Whichever of these
two lands second should re-verify the recovery path against the other's
behaviour — I don't think they conflict textually, but the timing assumption is
shared.

## Suggested merge checklist

- [ ] Wrap the `GateReleased` write in `RetryOnConflict` and return the error
- [ ] Use a live `APIReader` list on the empty-`RestoredPodName` path
- [ ] Add skew tolerance to the `mismatch-since` comparison
- [ ] Confirm missing-annotation handling on upgrade from an older controller
- [ ] Add the two tests above
- [ ] Squash the duplicated commits

kyber775:
Addressed Round 3 review feedback in commit `9c36a89`:

1. **Durable `GateReleased` status write via `RetryOnConflict`:**
   In `pod_gate_controller.go`, wrapped the `GateReleased = true` status update in `retry.RetryOnConflict(retry.DefaultRetry, ...)` and returned the error if retries are exhausted (`return ctrl.Result{}, err`). This prevents concurrent status writers from dropping the marker and guarantees it is durably committed before the reconcile succeeds.
2. **Live `APIReader` list on empty `RestoredPodName`:**
   In `checkConsumerPodExists`, when `RestoredPodName == ""` (legacy PMJs), the reconciler now performs a live namespace list using `r.apiReader().List()` and filters in memory by `pod-migration.gke.io/assigned-pmj`. Bypassing the informer cache ensures cache lag cannot falsely report a living consumer as deleted or trigger a duplicate restore.
3. **Clock skew tolerance on `mismatch-since`:**
   Added a `clockSkewTolerance = 10 * time.Second` buffer to `mismatchGracePeriod` in `podmigrationjob_controller.go`, absorbing wall-clock drift between HA replicas and preventing premature expiration during leader failovers. Confirmed that missing annotations on upgrade are initialized on first observation without spurious fast-fails.
4. **Regression tests added:**
   - `TestPodGate_GateReleased_ConflictRetry`: simulates a 409 Conflict via `interceptor.Funcs` on the status subresource update, asserting `RetryOnConflict` recovers and persists `GateReleased = true`.
   - `TestPodGate_EmptyRestoredPodName_UsesLiveReader`: asserts that with an empty informer cache and a pod present on the live API via `APIReader`, recovery does not fire and the candidate pod remains gated.
   - Updated `TestPodMigrationJobReconciler_Restoring_UIDMismatch_PersistedTimeout` to verify the skew-adjusted expiration window.
5. **Squash on merge:**
   Recommending squash-and-merge to squash the branch into a single atomic commit on `upstream/main`.

ahahadelyaly:
# Review comment — PR #36 (Round 3)



---

Re-reviewed Round 3 against the merge base. The live-reader fix is correct and
the self-heal you added is a genuinely good addition. One ordering change I'd
like before merge — it's about five lines.

## Confirmed fixed

**C2 — cache-only existence check.** The empty-`RestoredPodName` path now goes
through a live `apiReader()` list with in-memory filtering rather than the
index-backed cached `List`. That's the right resolution: `APIReader` can serve
`List`, it just can't use a field index, so filtering in memory keeps the check
authoritative without giving up recovery. The two obvious alternatives — assume
the pod is alive, or return an error and requeue — would have traded corruption
for a permanent strand or an infinite thrash respectively. You avoided both.

**Zero-value annotation handling.** The empty-string bypass means a CR written by
an older controller doesn't parse into a zero time and get treated as infinitely
old. That was the upgrade landmine; it's closed.

**The `GateReleased` self-heal is a good call.** Nobody asked for it, and it
meaningfully reduces the blast radius of a lost sentinel write. Worth keeping
regardless of the change below.

## Blocking: swap the operation order

The sentinel currently commits *after* the pod is ungated:

```go
r.removeGate(pod)                       // irreversible
if err := r.Update(ctx, pod); err != nil { ... }

if !job.Status.GateReleased {           // durable record, second
    err := retry.RetryOnConflict(...)
}
```

So if the sentinel write fails, you're left with a **running, ungated pod and no
record of it**. The operation fails open.

To be clear about severity, since the self-heal changes the picture: this is
**not** permanent loss. The PMJ reconciler's self-heal catches
`!job.Status.GateReleased && !podHasMigrationGate(replacementPod)` and stamps it,
normally within a 2s requeue. So the exposure is a narrow race, not a guaranteed
bug — if the consumer pod dies inside that window, stranded recovery reads
`GateReleased == false`, concludes the pod died before it ever left the gate, and
re-assigns the snapshot to a replacement. Two pods, one checkpoint.

Two things make that window wider than it first looks:

1. The `RetryOnConflict` closure re-`Get`s through the **cached** client. On a
   lagging informer it can burn its whole backoff against the same stale
   `resourceVersion`.
2. When it exhausts and returns the error, the requeue is a **dead end** — on the
   next pass the gate is already gone, `gateIndex == -1` early-returns, and the
   sentinel write is never retried. The self-heal is doing all the real recovery
   work here; the retry loop contributes less than it appears to.

Reordering removes the window entirely rather than shrinking it:

```go
// 1. Durable record first. If this fails, the pod is still gated — it cannot
//    restore, and the requeue retries safely.
if !job.Status.GateReleased {
    patch := client.MergeFrom(job.DeepCopy())
    job.Status.GateReleased = true
    if err := r.Status().Patch(ctx, job, patch); err != nil {
        return ctrl.Result{}, err
    }
}

// 2. Then the irreversible step.
r.removeGate(pod)
if err := r.Update(ctx, pod); err != nil {
    return ctrl.Result{}, err
}
```

Now the failure mode is a gated pod *with* a record — it can't restore, and retry
converges. Fails closed. Using `Patch` here also sidesteps the cached-`Get`
problem and avoids clobbering sibling status fields that a concurrent writer may
be touching.

Keep the self-heal as the backstop; it's still worth having for the case where
the pod is ungated by some path this reconciler didn't drive.

## Non-blocking

**Clock skew is mitigated, not cured.** The 10s tolerance makes the common case
safe, but it's still `time.Since()` against a wall-clock string stamped by
whichever replica held the lease. Fine to ship; worth a follow-up issue to move
to an API-server-generated timestamp rather than comparing across machine clocks.

**Please squash before merge.** The branch still carries the merge commit and
three duplicated commit pairs. I checked and the net tree is coherent — nothing
double-applied, no earlier review fix dropped — so this is purely `git blame`
hygiene, not correctness. An interactive rebase down to atomic commits would do
it.

**Rebase after the sibling PR lands.** This is still based on the pre-`#38`
commit. The other in-flight PR touches a disjoint file set (the `PodMigration`
controller and deploy manifests, versus this PR's PMJ and pod gate controllers),
so no textual conflict is expected — but please re-run `make verify-generate`
after the rebase, since this PR regenerates the CRD and the freshness gate will
catch any drift.

Happy to re-review as soon as the reorder is in.

kyber775:
Addressed Round 3 review feedback in commit `0f08c4b`:

1. **Reordered gate release (fails closed):**
   In `pod_gate_controller.go`, reordered the sequence so `GateReleased = true` is durably patched on PMJ status via `r.Status().Patch(ctx, job, patch)` **before** `r.removeGate(pod)` and `r.Update(ctx, pod)`. If the status patch fails or conflicts, the reconciler returns the error immediately while the candidate pod remains held in `Pending` behind `gke.io/pod-migration-gate`. The operation now fails closed, eliminating the window where an ungated, executing pod could exist without a durable status record. Using `Patch` also sidesteps cached-`Get` resourceVersion conflicts and prevents clobbering sibling status fields touched by concurrent reconcilers.
2. **Backstop self-healing retained:**
   Preserved the self-healing check in `PodMigrationJobReconciler`, which stamps `GateReleased = true` whenever an un-gated replacement pod is observed during `Restoring` reconciles.
3. **Rebased onto `upstream/main` post-PR #35:**
   Cleanly rebased against `upstream/main` following the merge of PR #35. Regenerated CRD manifests via `make manifests` and verified zero drift via `make verify-generate`.
4. **Regression test added:**
   Added `TestPodGate_GateReleased_StatusPatchFailure_PodRemainsGated` simulating a status patch failure via `interceptor.Funcs.SubResourcePatch`: verifies that the candidate pod retains its scheduling gate (fails closed) when the status patch errors, and verifies that the subsequent retry succeeds, persists `GateReleased = true`, and releases the gate.
5. **Full test suite clean:**
   All unit test suites across `controller/internal/...` pass 100%.

PR #39, opened 2026-09-18
fix(controller): index VolumeAttachments by PV name for O(1) detach checks (Closes #24)

### LINKED ISSUE #24: PV detach wait lists all VolumeAttachments cluster-wide every 3 seconds per migration
## Problem

While a PMJ with PVs waits for volume detach in the `Evicting` phase, every reconcile lists **all VolumeAttachments in the cluster** and scans them, on a 3-second requeue loop. Work is O(cluster attachments) per PMJ per tick; with N concurrent migrations of PV-backed pods on a storage-heavy cluster this becomes a meaningful cache scan load, and it is unnecessary: the set of PVs of interest is already known and small.

## Where in the code

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/podmigrationjob_controller.go#L426-L452

```go
vaList := &storagev1.VolumeAttachmentList{}
err = r.List(ctx, vaList)            // cluster-wide, every 3s per waiting PMJ
...
for _, va := range vaList.Items {    // scan all attachments
```

## Suggested fix

Index VolumeAttachments by PV name at manager setup and query only the PVs the job is waiting on:

```go
// cmd/main.go
mgr.GetFieldIndexer().IndexField(ctx, &storagev1.VolumeAttachment{}, "spec.pvName",
    func(o client.Object) []string {
        va := o.(*storagev1.VolumeAttachment)
        if va.Spec.Source.PersistentVolumeName == nil { return nil }
        return []string{*va.Spec.Source.PersistentVolumeName}
    })

// reconciler
for _, pv := range job.Status.PVsToDetach {
    vaList := &storagev1.VolumeAttachmentList{}
    if err := r.List(ctx, vaList, client.MatchingFields{"spec.pvName": pv}); err != nil { ... }
    for _, va := range vaList.Items {
        if va.Status.Attached { stillAttached = true }
    }
}
```

Optionally also replace the 3-second poll with a `Watches` on VolumeAttachment mapped through the same index back to waiting PMJs, keeping a slow safety requeue.

kyber775:
## Summary

Indexes `VolumeAttachment` objects by persistent volume name (`spec.source.persistentVolumeName`) in `RegisterFieldIndexes`, replacing cluster-wide unindexed cache scans during PMJ `Evicting` volume detach wait with targeted $O(1)$ indexed field queries per target PV.

## Why This Change

In `PodMigrationJobReconciler` during the `Evicting` phase (step 4.3), when a PMJ has persistent volumes to detach (`job.Status.PVsToDetach`), every reconcile execution listed all `VolumeAttachment` resources across the entire cluster via unindexed `r.List(ctx, vaList)` and performed a full nested slice scan every 3 seconds per waiting PMJ.
In storage-heavy clusters with hundreds or thousands of volume attachments and concurrent migrations, this imposed significant cache scan load and memory allocation churn on every 3-second requeue loop. Because the set of target PV names is already known and small, indexing `VolumeAttachment` by PV name reduces cache scan work from $O(\text{cluster attachments})$ to $O(1)$ per target PV.

## What Changed

- **`controller/cmd/main.go`**:
  - Updated `RegisterFieldIndexes` setup comment to document registration of both `PodAssignedPMJIndex` and `VolumeAttachmentPVIndex`.
- **`controller/internal/controller/indexers.go`**:
  - Added `VolumeAttachmentPVIndex` constant (`"spec.source.persistentVolumeName"`).
  - Added `VolumeAttachmentPVIndexValue` indexer function with nil-safe and empty-string extraction of `va.Spec.Source.PersistentVolumeName`.
  - Added typed nil safety check to `PodAssignedPMJIndexValue`.
  - Registered `VolumeAttachmentPVIndex` on `storagev1.VolumeAttachment` in `RegisterFieldIndexes`.
- **`controller/internal/controller/indexers_test.go`**:
  - Added `TestVolumeAttachmentPVIndex_ListsOnlyMatchingAttachments` verifying that indexed queries return only attachments matching the queried PV while filtering out unrelated, nil-PV, and empty-PV attachments.
  - Added `TestIndexValues_NilSafety` verifying nil-safety against untyped and typed nil interfaces.
- **`controller/internal/controller/podmigrationjob_controller.go`**:
  - Replaced cluster-wide `r.List(ctx, vaList)` with per-PV indexed queries `r.List(ctx, vaList, client.MatchingFields{VolumeAttachmentPVIndex: targetPV})`.
  - Added early-break optimization as soon as an active attachment is detected for any target PV.
- **`controller/internal/controller/podmigrationjob_controller_test.go`**:
  - Updated `TestPodMigrationJobReconciler_Evicting_Success` fake client to include `VolumeAttachmentPVIndex`.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_StillAttached_Requeues`.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_Detached_TransitionsToRestoring`.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_MultiplePVs_OneStillAttached`.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_ListError_ReturnsError`.

## Context

Closes #24

## Testing

- **Sabotage Verification**:
  - Sabotaged `VolumeAttachmentPVIndexValue` to return `nil`; verified that `TestVolumeAttachmentPVIndex_ListsOnlyMatchingAttachments` and `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_StillAttached_Requeues` fail.
- **Unit Test Suite**:
  - `cd controller && go test -race -v -count=1 ./...` (all suites pass with zero race conditions).
- **Manifest Verification**:
  - `cd controller && make verify-generate` (code generation and deploy assembly clean).

## Live validation

Not live-tested on GKE cluster: this is internal controller-runtime client-side cache indexing and field querying logic, fully validated via unit tests, fake client indexers, and sabotage testing.

## Self-Review

Isolated pre-PR review conducted via subagent `deleg_e3acf3d4`:
- **Finding 1 ([CRITICAL])**: Typed nil interface dereference panic in `VolumeAttachmentPVIndexValue`. If `client.Object` wraps a typed nil pointer `(*storagev1.VolumeAttachment)(nil)`, type assertion succeeds with non-nil `ok` but nil pointer dereference on `va.Spec`.
  - *Disposition*: Fixed; added `va == nil` guard in `VolumeAttachmentPVIndexValue` and `pod == nil` guard in `PodAssignedPMJIndexValue`, covered by `TestIndexValues_NilSafety`.
- **Finding 2 ([WARNING])**: Detachment loop lacks a timeout ceiling.
  - *Disposition*: Preserves existing controller behavior. Adding a detachment timeout or status condition changes CRD/status semantics and is tracked as a separate follow-up enhancement.
- **Finding 3 ([WARNING])**: PV matching evaluates cluster-wide attachments without node scoping for multi-attach/RWX volumes.
  - *Disposition*: Preserves existing controller behavior. Scoping attachments to `OriginNodeName` requires adding node tracking fields to `PodMigrationJobStatus` and is tracked as a separate follow-up enhancement.
- **Finding 4 ([SUGGESTION])**: Stale cache index comment in `controller/cmd/main.go`.
  - *Disposition*: Fixed; updated comment to accurately document both indexes.

## Risk & Rollout

- **Blast radius**: Scoped strictly to the volume detachment wait check in `PodMigrationJobReconciler` during the `Evicting` phase.
- **Rollback**: Standard controller deployment rollback; no CRD schema or RBAC modifications.


ahahadelyaly:
Hi, I'm working on pod migration guard rail this week so I'll hold the review a few days while I'm finishing wiring up the guardrail

bnaylor (approved):
Round 1, at e78ad4f.  This is the right fix for #24 and it's the right size: one indexed cached List per PV instead of a cluster-wide scan every 3s per migration, registered at boot alongside the assigned-pmj index, RBAC already covered.  I traced the new loop against the old one and the short-circuit semantics are identical.  The tests are real - the still-attached and detached cases falsify each other, the multi-PV case exercises the outer break, and the interceptor test pins error propagation.  Build, vet, tests, verify-generate clean.

Two notes, neither blocking:

- The wait still matches by PV name only, so a destination-node attachment for the same PV would hold it just like the old code.  You called this out yourself in the PR body; please file it as an issue so it doesn't live only there.  The PMJ doesn't record the origin node today, so it's a small status addition when someone picks it up.
- Registering the index moves the VolumeAttachment informer from lazy to boot-time, which means the readyz startup sync from #30 now waits on a VA list even on an idle controller.  VAs are bounded by attached PVs, so fine, just naming the change.

Approving.

PR #41, opened 2026-09-21
feat(controller): detect failed snapshot restores and fall back to a cold start

### LINKED ISSUE #15: No cold-start fallback when restore crashes: CompleteRestore failures leave pods in permanent CrashLoopBackOff
## Problem

When the gVisor restore itself crashes, the replacement pod is stranded in permanent CrashLoopBackOff. There is no cold-start fallback on this path (the restore-skip path falls back; the restore-crash path does not). Typical signature:

```
OCI runtime restore failed: starting container: ... failed to load kernel:
vfs.CompleteRestore() failed: failed to complete restore for filesystem type "9p":
failed to walk "temp-52.rdb" of type 0100000 in mount "main:/data": no such file or directory
```

Any workload that modifies its writable layer between checkpoint and restore can hit this. In an eviction test of 10-replica StatefulSets on GKE 1.36.3-gke.1537000, four mainstream databases failed restore this way at 100% rate under normal write activity, no induced churn: MongoDB (`collection-*.wt`), PostgreSQL (`pg_wal`), MySQL, RabbitMQ (`mnesia`). Each stranded pod then also resisted deletion (`KillPodSandbox ... signaling process in container ... PID 0: no such process`), blocking its StatefulSet ordinal until `kubectl delete pod --force`.

So today the practical outcome for the whole disk-writing database class is: migration converts a routine drain into a permanent, manually-recovered outage, which is strictly worse than a cold eviction.

## Where in the code

After the gate controller injects `podsnapshot.gke.io/ps-name` and releases the gate, no component watches the replacement, so the crashloop is never detected:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/internal/controller/pod_gate_controller.go#L116-L131

## Suggested fix

1. Watch replacement pods the controller annotated. On restore start failure, degrade to an evented cold start:

```go
// in a replacement-pod reconcile
for _, cs := range pod.Status.ContainerStatuses {
    t := cs.LastTerminationState.Terminated
    if t != nil && t.ExitCode == 128 && strings.Contains(t.Message, "restore") && cs.RestartCount >= 2 {
        delete(pod.Annotations, "podsnapshot.gke.io/ps-name")
        delete(pod.Annotations, "pod-migration.gke.io/assigned-pmj")
        recorder.Event(pod, corev1.EventTypeWarning, "RestoreFailedColdStart",
            "restore crashed repeatedly; falling back to a cold start")
        return r.Delete(ctx, pod) // workload controller recreates it clean
    }
}
```

If the pod then sticks in Terminating past a deadline (the sandbox-kill failure above), escalate to deletion with grace period 0 and surface a warning event, matching the manual runbook but automated and visible.

2. Prevention at admission: the repository already ships an optional policy (`patches/gke-pod-snapshot-admission-webhook.yaml`) to reject incompatible workloads. Given a measured 0/40 restore success across four major databases, consider making a policy that covers known disk-writing database images part of the default install until the underlying checkpoint/restore behavior handles writable-layer drift.

ahahadelyaly:
### Problem

When a snapshot cannot be restored onto a replacement pod, the container runtime
fails to start the container and **the kubelet never retries**. The pod is wedged
permanently, and the migration hangs until the PMJ is garbage collected 30 minutes
later. There is currently no detection and no recovery: the workload is simply down.

Closes #15

### Solution

```mermaid
flowchart TD
    A["PMJ Restoring, replacement pod observed (UID matched)"] --> R{"Pod already Ready?"}
    R -->|"yes"| C["Step 7/8: cold-start event check -> Succeeded"]
    R -->|"no"| B{"Container start failure in CURRENT state?<br/>Terminated(StartError,128) or<br/>Waiting(RunContainerError) + LastTerm(StartError,128)"}
    B -->|"no"| P["Step 6: throttled cold-start event probe"]
    B -->|"yes, known signature"| L{"Live APIReader re-Get<br/>still FailureFatal?"}
    L -->|"no (recovered)"| Q["Requeue 2s, no status write / no delete"]
    L -->|"yes"| D["Phase=Failed + RestoreCrashed<br/>persist status (clear Unrecognized if set)"]
    D --> E["Delete pod (UID precondition)"]
    E --> F["Workload controller recreates it cold<br/>PMJ is Consumed, so no re-adoption"]
    B -->|"yes, unknown signature"| G["Metric + Warning event + condition<br/>NO action taken"]
    G --> H["5m ceiling (also checks live for fatal crash first)"]
```

Three ordering and safety guarantees worth reviewing closely:

- **Live `APIReader` confirmation before the status write (and delete).** The 2s reconcile loop classifies against the informer cache, which is cheap; if it sees `FailureFatal`, we re-Get and re-classify via `APIReader` before writing `Phase=Failed` or deleting the pod (threading the live pod into the deleter so it costs only a single live GET).
- **Status is persisted before the pod is deleted.** If we deleted first and then crashed, the PMJ would sit in `Restoring` with its recorded pod gone and no record of why.
- **Deletion is idempotent, retried, and runs before the 5m ceiling concludes `SucceededWithoutRestore`.** Because the phase is already `Failed`, a failed `DELETE` (a throttled request during a large evacuation is the realistic case) can no longer be retried by the `Restoring` branch. Both the first attempt, the `Failed`-phase completion path, and the post-5m-ceiling crash check call the same helpers (`concludeRestoreCrash` / `ensureCrashedReplacementPodDeleted`) so they cannot drift apart, and the deleter refuses to delete a same-named successor by asserting the recorded UID.

### How the detection signature was derived

The signature was **not** inferred from upstream source. It was captured by replaying
real restore failures on a live GKE cluster with a gVisor node pool. Full methodology
and raw captures are in an internal doc; the load-bearing results are reproduced here
so this PR is self-contained.

| Scenario | Current `state` | `lastState.terminated.reason` | Exit | `restartCount` | Distinctive message |
|---|---|---|:---:|:---:|---|
| Happy path (S1) | `Running` | — | — | `0` | `Successfully restored the pod from PodSnapshot` |
| **Restore failure (S2)** | `Waiting(RunContainerError)` or `Terminated(StartError)` | `StartError` | **`128`** | **`0`, never increments** | `failed to start containerd task ... OCI runtime restore failed` |
| Missing checkpoint (S4) | `Running` (cold) | — | — | `0` | `GKEPodSnapshotting: falling back to a cold start ... does not exist` |
| Application bug (S5) | `Waiting(CrashLoopBackOff)` | `Error` | `1` | **increments** | `BackOff: Back-off restarting failed container` |
| **Recovered pod (S6)** | **`Running` (Ready)** | **`StartError` (stale)** | **`128`** | `>= 1` | *(same as S2 — persists in `LastTerminationState` forever)* |

Restore failures (S2), application bugs (S5), and recovered pods (S6) are **100% separable**:
- **S2 vs S5:** A restore failure never reaches the guest (the sandbox itself fails to construct, so the CRI reports `StartError` with exit `128`), whereas an application bug means the sandbox constructed fine and the process then exited on its own terms.
- **S2 vs S6:** A container that failed to start once and later started keeps `StartError/128` in `LastTerminationState` forever. The classifier therefore derives its verdict **strictly from the container's current state** (`Running` / `Completed` containers are skipped; `LastTerminationState` is consulted only when the container is currently parked in `Waiting(RunContainerError)`), and step 5 is additionally gated on `!isPodReady`.

> [!WARNING]
> **Please do not add a `restartCount` threshold to this detector.**
> On a restore failure the kubelet never restarts the container, so `restartCount` is
> pinned at `0` forever. The intuitive implementation — *"trigger fallback once
> `restartCount >= 3`"* — would **never fire**, shipping a silent no-op that looks
> correct in review. There is a regression test (`TestClassify_IgnoresRestartCount`)
> specifically to break anyone who adds one.

**Only empirically observed signatures are actioned.** An earlier iteration of this
work carried a `"does not match"` pattern; it was removed because it was never
observed in any replay, and acting destructively on an invented signature inverts the
report-first design. An unrecognised start failure is reported via metric, Warning
event and condition (`UnrecognizedRestoreCrash`, retracted if the pod later recovers),
but never acted on — so if a real checkpoint-mismatch message exists in the wild, it
reaches us as monitoring signal and can then be added *with evidence*, rather than us
guessing now.

### Why a new `internal/restore` package

Pod migration varies along **two independent dimensions**:

| Dimension | Concern | Abstraction |
|---|---|---|
| Snapshot orchestration | requesting a checkpoint, learning its state | `snapshot.Provider` (already existed) |
| Checkpoint/restore engine | what freezes/thaws the process, and how it reports failure | `restore.Engine` (**new**) |

These must not be conflated, because GKE Pod Snapshots is expected to gain CRIU
support — meaning the orchestration layer can stay fixed while the engine underneath
changes.

The split that keeps this cheap: the kubelet start-failure **shape** is a kubelet+CRI
property, identical for every engine, so it lives once in `restore.startFailures`. An
`Engine` contributes only its **message vocabulary**. The reconciler keeps all policy
(conditions, deletion, metrics, events).

Concretely: **adding CRIU later is a new `Engine` plus one line in `DefaultEngines()`,
touching neither `podmigrationjob_controller.go` nor `snapshot.Provider`.** That
invariant is the acceptance test for whether the seam is in the right place.

The repo targets gVisor only today — `DefaultEngines()` returns exactly one engine.
Nothing speculative is enabled.

### Job `backoffLimit` & `podFailurePolicy` (`README.md` + sample manifests)

A crash-deleted Job pod terminates `Failed` with exit code **`128`** (`StartError`).
Because the status mutating webhook only rewrites pods transitioning to `Succeeded`
during `Snapshotting`/`Evicting`, the crashed replacement pod never receives the
synthetic `137` exit code — so without an additional `podFailurePolicy` match on `128`,
every cold-start fallback silently consumes one `backoffLimit` retry.

`README.md` §3 and both Job manifests (`pm-go-job.yaml`, `pm-node-job.yaml`) are
updated to `Ignore` exit codes `[137, 128]`, with the customer precondition stated
explicitly (the application itself must not exit `128`) and a collapsible note
explaining why `DisruptionTarget` cannot be used (plain `DELETE` never sets it, and
routing through the Eviction API can be refused by a PDB because the crashed pod is
not Ready).

### ⚠️ Interaction with #36 / Issue #25 — please sanity-check this

Rebasing onto #36 surfaced a real semantic interaction, not just a textual conflict.

`pod_gate_controller.go` now recovers a stranded PMJ when its recorded consumer pod
has vanished, but deliberately scopes that to `Phase == Restoring && !GateReleased`.

**This fallback can never produce that state**, because it persists `Phase=Failed`
*before* deleting, and the crashed pod had necessarily been un-gated already (it
started, then failed to restore). So the two features own disjoint states, and the
`Phase` and `GateReleased` guards are independent and both load-bearing.

One test fixture in this PR was corrected accordingly — it had been modelling
`Restoring && !GateReleased`, which is unreachable from this code path and is now
Issue #25's territory. The reasoning is written into the test so a future cleanup
does not "simplify" one of the guards away.

### Out of scope

- **Missing checkpoints.** `gke-pod-snapshots` intercepts before containerd
  initialises the sandbox and cold-starts natively, so no controller-side handling is
  needed.
- **Node/runtime version pinning.** Not solvable at this repo's level.

### Follow-ups (to be filed)

- Move the cold-start fallback detection behind `snapshot.Provider` — it currently
  hardcodes GKE event vocabulary in the reconciler, completing the dimension-1 seam.
  Several of its match strings are also unverified and should be audited.
- Add a CRIU `restore.Engine` once GKE Pod Snapshots ships CRIU.

### Observability

- `pod_migration_restore_crash_fallback_total`
- `pod_migration_restore_crash_unmatched_total`
- Warning events `RestoreCrashFallback` / `UnrecognizedRestoreCrash` (event emission
  is once-only, guarded by the condition transition, so a 2s requeue loop cannot spam)
- Conditions `RestoreCrashed` / `UnrecognizedRestoreCrash` (retracted to `False` if the pod recovers)

### Testing

`gofmt` clean · `go vet` clean · `make verify-generate` no drift · `go test -race -count=1 ./...` all green.

New coverage includes:
- Fatal path end-to-end; app-crashloop negative control; unmatched start error does not delete; event emitted exactly once; delete retried after a transient failure (mutation-verified via `interceptor.Funcs`); completion path spares a same-named successor; re-adoption blocked at both guard layers; engine-seam test using a fake CRIU engine; engine-ordering test.
- **Round 1 regression tests (verified by negative control — reverting each fix and confirming failure):**
  - `TestClassify_IgnoresStaleTerminationHistory` (5 cases × container/initContainer: Running + stale `StartError/128`, `CrashLoopBackOff` + stale history, `Completed` exit 0 + stale history, `StartError` with non-128 exit, `RunContainerError` with non-128 history)
  - `TestPodMigrationJobReconciler_Restoring_RecoveredReadyPodIsNotDeleted` (K1/K2)
  - `TestPodMigrationJobReconciler_Restoring_CrashAfterCeilingStillRunsFallback` (B1)
  - `TestPodMigrationJobReconciler_Restoring_StaleCachedCrashIsNotActedOn` (B2)
  - `TestPodMigrationJobReconciler_Restoring_UnrecognizedConditionClearedOnSuccess` & `NoVacuousUnrecognizedCondition` (B4)


kyber775 (changes requested):
## Review: PR #41 - feat(controller): detect failed snapshot restores and fall back to a cold start

### Summary & Verdict
**Verdict:** `REQUEST_CHANGES`

The architectural framing of this change is excellent: decoupling the checkpoint/restore engine signature taxonomy (`internal/restore`) from the controller reconciliation policy, persisting `Phase=Failed` *prior* to pod deletion, and sharing the idempotent `ensureCrashedReplacementPodDeleted` helper between the initial transition and the `Failed`-phase retry loop are all well-designed Day-2 SRE patterns.

However, there is a critical bug in `startFailures` and its ordering within `Reconcile`: **unconditional inspection of `cs.LastTerminationState` causes healthy, currently running/ready replacement pods to be classified as fatal restore crashes and deleted**.

---

### [CRITICAL] `controller/internal/restore/classifier.go:96-108` & `controller/internal/controller/podmigrationjob_controller.go:729`
**Historical `LastTerminationState` matches running/ready containers, causing live pods to be deleted**

#### Problem
In `startFailures(pod *corev1.Pod)`:
```go
		var startFail *corev1.ContainerStateTerminated
		for _, term := range []*corev1.ContainerStateTerminated{cs.State.Terminated, cs.LastTerminationState.Terminated} {
			if term == nil || term.ExitCode != startFailureExitCode {
				continue
			}
			if term.Reason == "StartError" || runContainerError {
				startFail = term
				break
			}
		}
```
`cs.LastTerminationState.Terminated` records historical termination state from an earlier execution.
If a container or init container experienced a start error on its first attempt (exit code 128 / `StartError`), but kubelet subsequently retried and the container is now **`Running`** (`cs.State.Running != nil`) or **`Completed`** (`cs.State.Terminated.ExitCode == 0`), `cs.LastTerminationState.Terminated` retains the historical exit code 128 and `StartError`.

Because `startFailures` iterates over `[]*corev1.ContainerStateTerminated{cs.State.Terminated, cs.LastTerminationState.Terminated}` without checking whether the container is currently `Running`:
1. If the historical message contains an engine signature (e.g. `"OCI runtime restore failed"`), `Classify` returns `FailureFatal` on a **running** container.
2. In `podmigrationjob_controller.go:730`, `case restore.FailureFatal` immediately fails the migration and **deletes the live, running replacement pod**.
3. If the historical message does not match a signature, it returns `FailureUnrecognized`, which writes `ConditionUnrecognizedRestoreCrash = True` and returns `ctrl.Result{RequeueAfter: 2 * time.Second}, nil`, trapping the controller in an infinite requeue loop even when the pod is healthy and Ready.

#### Empirical Reproduction
We verified this against the PR code with a test fixture where the container is `Running` and the pod is `Ready`, but `LastTerminationState` has a prior `StartError`:
```go
// Output when reconciled through PodMigrationJobReconciler:
// BUG CONFIRMED: Ready pod was deleted! err=pods "test-pod" not found
```
Similarly, if a container is currently in `CrashLoopBackOff` (`cs.State.Waiting.Reason == "CrashLoopBackOff"`), `startFailures` also matches `cs.LastTerminationState.Terminated`, mistaking an application crash loop for an active start failure.

#### Fix
`cs.LastTerminationState.Terminated` is only relevant when the container is currently parked in `Waiting` with `Reason == "RunContainerError"`.
1. If `cs.State.Running != nil`, the container is running and is NOT in a start failure.
2. If `cs.State.Terminated != nil && cs.State.Terminated.Reason == "StartError" && cs.State.Terminated.ExitCode == startFailureExitCode`, it is a direct start failure.
3. If `cs.State.Waiting != nil && cs.State.Waiting.Reason == "RunContainerError"`, then inspect `cs.LastTerminationState.Terminated` for `Reason == "StartError"` and exit code 128.
4. If `cs.State.Waiting != nil && cs.State.Waiting.Reason == "CrashLoopBackOff"`, or if an init container completed with `ExitCode == 0`, skip.

---

### [CRITICAL] `controller/internal/controller/podmigrationjob_controller.go:729`
**Step 5 (Restore-crash detection) runs before checking if the replacement pod is already `PodReady`**

#### Problem
In `podmigrationjob_controller.go`, Step 5 (`restore.Classify`) runs unconditionally on lines 729-816 before Step 6 (`if !isPodReady(replacementPod)`) and Step 7 (`Pod is Ready`).

A replacement pod that is already `PodReady == ConditionTrue` has successfully initialized its sandbox, restored or cold-started, and is passing readiness probes. Checking for restore crashes on a pod that is already Ready:
- Inverts the state machine hierarchy (a pod cannot be both Ready and unrecoverably wedged in a restore crash).
- Exposes live traffic-serving pods to destructive fallback deletion.
- In the case of `FailureUnrecognized` (e.g. from an init container or sidecar artifact), Step 5 intercepts the loop and returns `requeue: 2s`, preventing the reconciler from ever reaching Step 7 to mark the migration `Succeeded`.

#### Fix
Gate restore-crash detection so it only executes when `!isPodReady(replacementPod)`. If `isPodReady(replacementPod)` is true, skip restore-crash detection and proceed directly to Step 7.

---

### [SUGGESTION] `controller/internal/restore/classifier.go:102`
**Tighten condition coupling in `startFailures`**

In:
```go
if term.Reason == "StartError" || runContainerError
```
When checking the termination state associated with `RunContainerError`, require both `runContainerError` AND `term.Reason == "StartError"` (or exit 128) rather than an `||` fallback, ensuring application exits (like `Error` / exit 1) under unusual status timing are never attributed to runtime start failures.

---

### [PASS] Design & Implementation Strengths
- **Engine Seam Architecture**: Decoupling the CRI/kubelet start error shape from engine-specific regex/substring vocabulary via `restore.Engine` is a very clean abstraction that will make future CRIU support trivial without mutating controller logic.
- **Ordered Status Write**: Writing `job.Status.Phase = Failed` and persisting it before deleting the crashed pod ensures state survivability across controller restarts.
- **Idempotent Deletion Helper with UID Precondition**: Sharing `ensureCrashedReplacementPodDeleted` between the initial transition and the `Failed`-phase retry loop, combined with `client.Preconditions{UID: &pod.UID}`, guarantees that transient delete failures (e.g. 429 throttling) can be safely retried without ever deleting successor pods.
- **Re-adoption Invariants**: Verified that both `FindUnassignedActivePMJ` and `PodGateReconciler` respect `Phase=Failed` and `GateReleased=true`, preventing successor pods from re-adopting the snapshot.

bnaylor:
Round 1, at 47ed902.  The shape of this is good: the engine seam is the right cut, status-before-delete is the right order and it's retried through one helper, the UID precondition is exactly what a same-name successor needs, and the PR body says what the code does, including the part about #36.  Build, vet, tests, verify-generate all clean here; go.sum untouched and the go.mod promotions are earned.

On Kyber's review: both criticals hold, and I reproduced them.  A Ready, Running pod with a leftover StartError/128 in `LastTerminationState` and a matching message gets deleted; with a non-matching message it sits in the 2s loop with UnrecognizedRestoreCrash set until the ceiling calls it SucceededWithoutRestore.  They're one fix: classify only the container's *current* state - `Terminated(StartError, 128)`, or `Waiting(RunContainerError)` with the StartError in the last termination - skip anything Running or Completed, and don't run step 5 at all on a Ready pod.  Kyber's suggestion to require both `runContainerError` and the StartError reason is part of the same tightening.  Two of your own tests are the ones to add: Running plus stale history, and Ready plus stale history.

Three of mine, the first reproduced:

1. **The ceiling runs before the crash check, and the ceiling doesn't delete.**  [Step 2](https://github.com/gke-labs/pod-migration/blob/47ed9029b89ba9d17422726a689967d7f3a3eb6a/controller/internal/controller/podmigrationjob_controller.go#L572-L576) returns SucceededWithoutRestore before the pod is even fetched, and for a Consumed PMJ nothing defers it.  So a recognised crash that's first observed after the five minutes - controller down, leader handoff, a backlog under exactly the evacuation load #15 was filed against - gets recorded as a success and the pod is left wedged, kubelet never retrying, forever.  That's #15 leaving by the side door.  Reproduced: RestoringStartTime ten minutes back, pod carrying the matching signature, one reconcile, phase SWR, pod still there.  When Consumed, fetch and classify before the ceiling, or make the ceiling branch classify and take the Fatal path first.  Either way the crash check has to be able to win.
   https://github.com/gke-labs/pod-migration/blob/47ed9029b89ba9d17422726a689967d7f3a3eb6a/controller/internal/controller/podmigrationjob_controller.go#L722-L730

2. **The delete is decided on the cache.**  `replacementPod` is a cached Get and so is the one inside [ensureCrashedReplacementPodDeleted](https://github.com/gke-labs/pod-migration/blob/47ed9029b89ba9d17422726a689967d7f3a3eb6a/controller/internal/controller/podmigrationjob_controller.go#L110-L138).  This same file already reaches for the APIReader in `hasColdStartFallbackEvent` for a decision that deletes nothing.  A live re-Get and re-classify immediately before the Delete is one call on a path that fires rarely, and it closes the cache-lag half of what Kyber found.  Cheap enough that I'd just do it.

3. **Job pods pay for the fallback out of backoffLimit.**  The README's Job section teaches an Ignore rule for exit 137, which is the status webhook's synthetic code.  A crash-deleted Job pod ends Failed with 128 and no rule covers it, so every fallback costs the Job one retry.  Not a code change - a paragraph in section 3 and a second rule (128, or a DisruptionTarget condition rule) in the sample manifest.  #15 names database Jobs specifically, so it belongs in this PR.

Smaller:

- UnrecognizedRestoreCrash is set True and never set back.  After the fix above a pod can recover and the PMJ can conclude Succeeded still wearing it; flip it False on the Succeeded path.
- The comments point at `pm-notes/technical_docs/...`, which isn't in this repo.  Point at the PR body table instead, since that's what a maintainer can actually read.
- "restore failed" as a generalised signature is a bet that no other StartError message contains those words.  The 128 gate bounds it and you've documented the tradeoff, so fine - but the first unmatched-counter spike is the moment to revisit it.

Kyber's fix, plus 1 and 2, the README paragraph for 3, and the two reproduction tests, and I approve.

ahahadelyaly:
Thanks both — every finding was real, and two of them were cases where this feature could have destroyed or stranded a healthy workload. All are fixed in `1ecced1`, pushed as a separate commit so you can review just the delta.

The branch is also rebased onto `36fc6f8` (PR #32), hence the force-push.

| # | Finding | Status |
|---|---|---|
| K1 | `startFailures` reads `LastTerminationState` unconditionally | Fixed — my bug |
| K2 | Step 5 runs before the `isPodReady` check | Fixed |
| K3 | Require reason **and** exit code | Fixed, folded into K1 |
| B1 | Ceiling concludes `SucceededWithoutRestore` before the pod is fetched | Fixed — worst of the set |
| B2 | Delete decided on cached reads | Fixed, with one deliberate deviation (below) |
| B3 | Job pods pay for the fallback out of `backoffLimit` | Fixed, exit code `128` |
| B4 | `UnrecognizedRestoreCrash` set True, never cleared | Fixed |
| B5 | Comments cite `pm-notes/…`, not in this repo | Fixed |
| B6 | `"restore failed"` is a bet | Agreed, no action |

---

### K1 + K3 — reading termination history unconditionally

@kyber775 is right and this is the serious one. A container that fails to start once and then comes up keeps `StartError/128` in `LastTerminationState` **forever**. The old loop walked `State.Terminated` and `LastTerminationState.Terminated` together and accepted either `Reason == "StartError"` **or** `runContainerError`, so a Running, Ready pod matched on the stale history entry alone — and got deleted. Where the message matched no signature it was worse in a quieter way: `Unrecognized` requeues every 2s forever.

The predicate now derives its verdict only from the container's **current** state, and consults history solely to recover the message for a container the kubelet has parked in `RunContainerError`. Both branches require reason **and** exit code, so the `||` is gone (K3).

Worth naming the root cause rather than just the symptom: detection was derived from a taxonomy built entirely from *crashed* pods. It has no row for a **recovered** one. That missing row is precisely what went unconsidered, and I've noted it in the code comment so the next person extending this knows to look for it.

### K2 — detection ran before the readiness check

Fixed; detection is now gated on `!isPodReady`. I also folded the cold-start event probe into the same block, since "what to do while the pod hasn't come up" is one condition and splitting it across two `isPodReady` tests is how they'd drift apart later.

### B1 — the ceiling beats the crash check

@bnaylor this is the one I'm most glad you caught. The ceiling sits above the pod fetch, and the gated-claimant deferral is guarded by `if !job.Status.Consumed` — so a **Consumed** PMJ went straight to `SucceededWithoutRestore` having never looked at its pod. Success recorded, nothing deleted, pod wedged forever.

And the trigger isn't exotic: a controller restart, a leader handoff, or an evacuation backlog is enough to push first observation past five minutes — and an evacuation backlog is the scenario #15 was filed for. The feature escaping by the side door in its own motivating case.

The ceiling now runs the fallback when it finds a crash, reusing the same `concludeRestoreCrash` helper as the main path so the two can't drift.

I kept the ceiling where it is rather than moving it below the pod fetch: step 3 requeues on `NotFound`, so relocating it would mean a replacement pod that never appears never times out. That trades one wedge for another.

### B2 — live reads before acting (one deliberate deviation)

Agreed. One change from how you worded it: you asked for the live re-Get **immediately before the Delete**, and I put it **before the status write** instead, so `Phase=Failed` is never persisted off stale data either.

I checked this isn't a cost I'm quietly imposing — it's the **same single live GET** either way, because the live pod is threaded into the deleter rather than re-fetched there. The one genuine addition is the Failed-phase retry path, which has no pod in hand and so does its own live GET per retry; that path only runs after a `DELETE` has already failed. The cached classify still carries every 2s tick; only the confirmation is live.

Happy to move it back to exactly where you specified if you'd rather.

### B3 — Job `backoffLimit`

Confirmed and fixed. To be precise about why nothing covered this: the status webhook only rewrites exit codes on pods transitioning to `Succeeded` during an active `Snapshotting`/`Evicting` migration, and it matches `pmj.Spec.TargetPodUID` — the crashed *replacement* pod satisfies none of those, so it never receives the synthetic `137`. It terminates with `reason: StartError`, exit **128**, uncovered.

`128` is added to the sample and to both verification-suite manifests (the README cites `pm-go-job.yaml` by path, so leaving them out of sync would have made the doc cite a file contradicting it).

**On the `DisruptionTarget` alternative** — I looked into it and I don't think it works, for three independent reasons:

1. The fallback issues a plain `DELETE`, and a plain `DELETE` never sets `DisruptionTarget`. That condition is only added by the Eviction API, kubelet node-pressure/graceful-shutdown, the taint manager, PodGC, and scheduler preemption. The rule would never fire.
2. Routing the fallback through the Eviction API to obtain the condition would make it PDB-aware — and the crashed pod is **not Ready**, so under the default `unhealthyPodEvictionPolicy: IfHealthyBudget` the eviction can be **refused**, wedging the exact pod the fallback exists to remove. (Different from #32, where PDB-safety was the goal.)
3. `onPodConditions` matches type and status but not `reason`, so even a synthesized condition would also swallow preemptions and node-pressure evictions — broader than a customer wants.

All three are written into a collapsible block in the README so the question is answered in the docs rather than only here.

The one thing the exit-code approach *does* ask of the customer is that their application never exits `128` itself, since `onExitCodes` can't additionally require `reason: StartError`. That's now stated as an explicit precondition rather than left implicit.

### B4 — clearing `UnrecognizedRestoreCrash`

Fixed. Particularly relevant after K1: a pod hitting an unrecognised start failure is deliberately *not* acted on, so it really can recover, and the condition would otherwise permanently mislabel a migration that went on to succeed. Retracted on all three conclusion paths, mirroring `clearEvictionBlockageConditions` — including its hygiene of not writing vacuous `False` conditions on jobs that never saw one.

### B5 / B6

`pm-notes/` references repointed at the taxonomy table in the PR body. On B6 — agreed, `"restore failed"` is a hedge rather than an observation; it's a strict substring-generalisation of the confirmed string so it can't miss what the observed one matches, and the per-signature provenance comments now say which is which.

---

### Tests

10 new cases. Rather than assert they cover the bugs, I reverted each fix and confirmed the tests actually fail:

- old predicate → 3 of 5 `TestClassify_IgnoresStaleTerminationHistory` cases fail (the two K3 cases pass either way — the old loop already required exit 128, so they guard the conjunction rather than K1 specifically)
- ceiling check disabled → `CrashAfterCeilingStillRunsFallback` fails with *"Ceiling recorded SucceededWithoutRestore for a crashed pod"*
- live confirm disabled → `StaleCachedCrashIsNotActedOn` fails on all five assertions

Every fixture in the stale-history test carries a **real gVisor signature** in its history, so nothing but the container's current state distinguishes it from a genuine crash.

`gofmt`, `go build`, `go vet`, `go test -race` all clean.

---

### One thing I'd like confirmed

@bnaylor — the rebase onto #32 sits alongside #36, and `TestPodGateReconciler_SuccessorPodOfConsumedPMJGetsColdStartBypass` needed `gateReleased: true` added to its fixture. My reasoning: #36's recovery deliberately claims `Restoring && !GateReleased`, and #15 can never produce that state, because it persists `Phase=Failed` before deleting and the crashed pod had necessarily already been un-gated. So the combination the old fixture described is unreachable. That reasoning is written into the test, but it's an assertion about *your* code's invariant rather than mine — worth a second pair of eyes.

kyber775 (approved):
## Review: PR #41 (Round 2, commit `1ecced1`) - feat(controller): detect failed snapshot restores and fall back to a cold start

### Summary & Verdict
**Verdict:** `APPROVE` (`[PASS]`)

Every finding from Round 1 has been addressed cleanly and verified with rigorous regression coverage:
- **K1 / K3 (Stale termination history):** `startFailures` now derives its verdict solely from the container's current state. Containers in `Running` or `Completed` are skipped immediately. `LastTerminationState` is consulted strictly when `cs.State.Waiting` is active with `Reason == "RunContainerError"`, and both the reason (`StartError`) and exit code (`128`) are required in conjunction.
- **K2 (Readiness gate ordering):** Restore-crash detection in Step 5 is now gated on `!isPodReady(replacementPod)`. Traffic-ready replacement pods fall straight through to Step 7 and conclude `Succeeded`.
- **B1 (Ceiling bypass):** The 5-minute safety ceiling now performs a live classification of consumed replacement pods before falling back to `SucceededWithoutRestore`, ensuring late-observed crashes under heavy controller backlogs trigger the cold-start fallback rather than stranding workloads.
- **B2 (Live API confirmation before deletion):** Destructive decisions are now confirmed against the API server via `classifyRecordedPodLive` prior to persisting `Phase=Failed` and issuing `Delete(client.Preconditions{UID})`, protecting against informer cache lag.
- **B3 (Job backoffLimit):** README Section 3.2 clearly documents the exit 128 failure mode, articulates why `DisruptionTarget` does not apply (plain `DELETE` does not synthesize the condition; Eviction API blocks on unready pods under `IfHealthyBudget`), and updates both `pm-go-job.yaml` and `pm-node-job.yaml`.
- **B4 (Unrecognized condition lifecycle):** `clearUnrecognizedRestoreCrash` now cleanly retracts transient unrecognized crash conditions upon success or resolution, without introducing vacuous conditions on uninvolved jobs.
- **Invariant Verification (#36 interaction):** Confirmed the `gateReleased: true` update in `TestPodGateReconciler_SuccessorPodOfConsumedPMJGetsColdStartBypass`. In a restore crash, the container has necessarily executed (and thus un-gated), so `GateReleased == true` and `Phase == Failed` both hold. Under #36, stranded PMJ recovery only applies if `Phase == Restoring && !GateReleased`. The two features operate on strictly disjoint states.

---

### Verification
- **Mergeability:** Clean against `upstream/main` (`git merge-tree` exited 0).
- **Unit & Race Tests:** `cd controller && go test -race -count=1 ./...` passed cleanly (10 new regression tests, all green).
- **Code hygiene:** `gofmt`, `go vet`, and manifest references are all consistent.

bnaylor (approved):
Round 2, at 1ecced1.  Everything landed and I re-ran the three reproductions from round 1: Ready pod with a stale matching history goes Succeeded and stays alive, the non-matching variant sets no condition and doesn't loop, and a live crash observed after the ceiling ends Failed with the pod deleted.  Rebase across #32 is clean, tests and verify-generate clean.  Your deviation on B2 is an improvement, not a deviation - the live read deciding before the status write is the right place, and threading the live pod into the deleter is what makes it the same one GET.  The README section on why not DisruptionTarget is the kind of thing I'd normally have to ask for twice.

On your #36 question: your reasoning holds, and Phase is the load-bearing guard, not GateReleased.  PodGate's check is `Phase != Restoring || GateReleased`, so Failed short-circuits it on its own, and `FindUnassignedActivePMJ` excludes Failed and Consumed independently.  One nuance worth knowing: the new ceiling path runs before #36's 4b self-heal, so if PodGate's own GateReleased write had failed, the ceiling can conclude Failed with GateReleased still false.  Harmless for the reason above, three separate gates block re-adoption, but it means "the fallback never produces !GateReleased" isn't quite true.  I'd have that fixture cover Failed with GateReleased false as well, to pin that Phase alone is sufficient.

One small thing, not blocking: the ceiling's own SucceededWithoutRestore block (lines 919-941) is still inline from #29 and doesn't clear UnrecognizedRestoreCrash, and that block is exactly where an unrecognised crash lands after the 2s loop.  Since the ceiling now takes a live verdict, clear it when that verdict is None (the pod recovered) and leave it standing when it's still Unrecognized, where it's telling the operator the truth.  Fold the block into the helper while you're there.

Heads up: Kyber's #46 and #45 both touch the same Restoring/ceiling code and `markSucceededWithoutRestore`; I've suggested #45 -> #41 -> #46 as the landing order on #46.

Approving.

PR #45, opened 2026-09-23
fix(controller): watch snapshot CRDs, relax polling, and patch status on PMJ (Closes #20)

### LINKED ISSUE #20: Snapshotting phase polls at 1s intervals and status Update conflicts spam ERROR logs; use watches and Patch

kyber775:
## Summary

Fixes #20: Under load, the `Snapshotting` phase polled GKE snapshot CRDs at 1s intervals with `RequeueAfter: 1 * time.Second` and wrote status updates with `r.Status().Update`. This caused:
1. Significant API server poll amplification scaling linearly with concurrent migrations.
2. Spurious optimistic-lock conflict errors (~11 `ERROR` log lines per migration) burying real operational errors.

## What Changed

1. **Watch GKE Snapshot CRDs:** Added controller-runtime `Watches` for `PodSnapshotManualTrigger` (`EnqueueRequestForOwner`) and `PodSnapshot` (`mapSnapshotToPMJ`) in `SetupWithManager`. State transitions from GKE now immediately wake up the reconciler via watch events.
2. **Relaxed Polling Interval:** Relaxed the in-progress requeue from 1s to a 30s safety backstop (solely guarding against missed informer watch events).
3. **Status Patching:** Switched all `PodMigrationJob` status updates from full `r.Status().Update` to `client.MergeFrom` patches via `patchStatus`.
4. **Clean Conflict Requeue:** Handled `apierrors.IsConflict` via `handleStatusError`, logging at `Info` level and requeuing silently (`ctrl.Result{Requeue: true}, nil`) rather than logging through `logger.Error`.
5. **Testing:** Added unit tests in `podmigrationjob_controller_test.go` for `mapSnapshotToPMJ` and `handleStatusError`.

## Testing

`go test -v ./...` passed 100% across all packages in `controller/`.


bnaylor:
Round 1, at 83e0dcd.  The three ideas are right: watch the snapshot objects instead of polling them, patch status instead of updating it, and stop hammering the queue every second.  The patch conversion is sound - merge patches carry only the fields this reconciler changed, so they can't clobber PodGate's writes, and I confirmed the ordering guarantees other PRs rely on (status before delete in #41) hold identically under Patch.  Build, vet, tests, verify-generate clean.

Two things I'd fix before merge:

1. **The watches make the GKE snapshot CRDs a hard boot dependency for the whole process.**  [SetupWithManager](https://github.com/gke-labs/pod-migration/blob/83e0dcdc103ca42d69107a4e1daa9fe752b00ce2/controller/internal/controller/podmigrationjob_controller.go#L932-L935) now watches two podsnapshot.gke.io kinds.  In controller-runtime v0.23.3 a missing kind is retried every 10s with an error log until the controller's cache-sync timeout (2 minutes by default), then Controller.Start fails, mgr.Start returns, and main exits - webhooks, PodGate, everything.  Before this PR those kinds were only touched lazily once a migration reached Snapshotting, so `make run` against kind, a non-GKE cluster, or any future envtest manager test all started fine.  Now they hang two minutes and crash-loop.  Either check the RESTMapper for the kinds at setup and skip the watches with a loud warning (falling back to the poll), or document the addon as a startup requirement in the README and in `make run`'s comment.  I'd do the first; the second is a footgun for every new contributor.

2. **The PodSnapshot mapper enqueues every Snapshotting PMJ in the namespace on every snapshot event.**  [mapSnapshotToPMJ](https://github.com/gke-labs/pod-migration/blob/83e0dcdc103ca42d69107a4e1daa9fe752b00ce2/controller/internal/controller/podmigrationjob_controller.go#L940-L960) matches on `SnapshotRef == ""` because the ref is only recorded at Ready.  In the 2k-pod single-namespace evacuation that #20 and #33 were about, that's an O(N) list per event and all N PMJs enqueued per event, times three to five events per snapshot.  No extra apiserver QPS (all cached), but the 50 workers spend their time re-running EnsureTrigger and CheckStatus for jobs whose snapshot didn't move, which delays the real transitions during exactly the scenario the poll relaxation was meant to help.  The PSMT already carries `status.snapshotCreated.name` well before Ready: record it into SnapshotRef (or a PendingSnapshotRef) as soon as CheckStatus sees it, add a field index on it next to the existing ones, and make the mapper an indexed lookup.  Then the fan-out collapses to the one PMJ that owns the snapshot.

Smaller:

- `client.MergeFrom` without the optimistic-lock option never returns Conflict, and `origJob` is never nil, so `handleStatusError`'s conflict branch is unreachable from any real call site.  The PR description's "clean conflict requeue" is the Patch doing the work, not the helper.  Fine, but say so, and the test for that branch is testing a path nothing takes.
- The 1s -> 30s test change passed because code and test moved together; the mapper test is the real one.
- Coordination: this, #41 and #46 all conflict on this file.  I've suggested on #46 that this lands first, then #41, then #46.  If that holds, #41's two new `Status().Update` sites (the Unrecognized condition and `concludeRestoreCrash`) need converting to `patchStatus` when it rebases - please leave that note on #41 so it isn't missed.

1 and 2 and I approve.

kyber775:
Rebased onto `upstream/main` (absorbing #41) and addressed both review items:

1. **CRD boot dependency eliminated via RESTMapper check:** `SetupWithManager` now queries `mgr.GetRESTMapper().RESTMapping` for `PodSnapshotManualTrigger` and `PodSnapshot` before adding watches. If the CRDs are not registered (e.g. envtest, local kind clusters, non-GKE setups), it logs an informative notice and skips the watches, falling back to 30s polling without hanging manager startup.
2. **Collapsed snapshot mapper fan-out to O(1) via field indexing:**
   - In `internal/snapshot/gke_provider.go`, `CheckStatus` now populates `SnapshotRef` as soon as `status.snapshotCreated.name` appears in the manual trigger (even while `PhaseInProgress`).
   - In `podmigrationjob_controller.go`, `PhaseSnapshotting` records and patches `job.Status.SnapshotRef` immediately when discovered in progress.
   - Registered `.status.snapshotRef` index in `internal/controller/indexers.go`.
   - `mapSnapshotToPMJ` now uses `client.MatchingFields{PMJSnapshotRefIndex: snapName}`, eliminating the O(N) namespace PMJ scan.
3. **Absorbed #41 & converted all remaining status writes to `patchStatus`:**
   - Threaded `origJob` into `concludeRestoreCrash` and converted to `patchStatus`.
   - Converted `UnrecognizedRestoreCrash`, `WaitingForVolumeDetach`, `ReplacementPodMismatch`, and `markSucceededWithoutRestore` to `patchStatus`.
4. **Cleaned up `handleStatusError` doc comment** noting that `client.MergeFrom` patches don't assert resourceVersion unless optimistic locking is configured.

Verified with `make verify-generate`, `go vet ./...`, and `go test -race -count=1 ./...`.

bnaylor:
Round 2, at 22b77fd.  Both items landed the way I asked: the RESTMapper check makes the watches optional, and the mapper is now an indexed lookup keyed on a SnapshotRef that's recorded as soon as the trigger names the snapshot, with the empty-ref fan-out gone entirely.  I checked that recording SnapshotRef during Snapshotting changes no downstream decision (PodGate's durable branch is phase-gated first).  The remaining `Status().Update` is the nil-orig fallback inside `patchStatus`, `concludeRestoreCrash` carries `origJob`, and the status-before-delete ordering from #41 is intact.  Build, vet, tests, verify-generate clean.

One thing left, and it's a one-liner:

1. **The skip fires on any error, not just a missing kind.**  [Lines 1308 and 1317](https://github.com/gke-labs/pod-migration/blob/22b77fd9f5d46ba4a6ae0521d55e621c7147e72d/controller/internal/controller/podmigrationjob_controller.go#L1306-L1320) test `err != nil`.  The lazy mapper does live discovery on that first call, so an apiserver that's unreachable for a second at boot - a control-plane upgrade, which is exactly when controllers restart - fails discovery, logs "CRD not registered", and that replica runs poll-only for its whole life with a message that points at the wrong cause.  A genuinely missing CRD satisfies `meta.IsNoMatchError`; a discovery failure doesn't.  Gate the skip on that, and return anything else from SetupWithManager so the pod restarts instead of quietly degrading.

Housekeeping: same self-merge pattern as #36 (52193c7 merges its own superseded twin), harmless under squash.  And #46 isn't rebased on this yet, so it'll conflict in the reconciler when this lands - expected, just don't be surprised.

Item 1 and I approve.

kyber775:
Addressed Round 2 feedback in f1bb042:

- **Gated watch skip on `meta.IsNoMatchError`:** Extracted `shouldWatchCRD` which specifically inspects `restMapper.RESTMapping` errors with `meta.IsNoMatchError`. If the CRD is genuinely not registered, it skips the watch and logs the poll fallback notice. If discovery fails due to transient apiserver unreachable / network errors, it propagates the error to fail `SetupWithManager` so the pod restarts cleanly instead of running poll-only.
- Added unit tests in `TestShouldWatchCRD` covering registered CRDs, `NoKindMatchError`, `NoResourceMatchError`, and discovery failure errors.

Verified with `go test -race -count=1 ./...` and `make -C controller verify-generate`.

PR #40, opened 2026-09-18
chore(build): tooling hygiene for setup-envtest, controller-gen caching, verify-generate trap, and version struct consolidation (Closes #34)

kyber775:
## Summary
Resolves non-blocking build, CI, and test tooling improvements identified during the PR #31 review cycles (Issue #34), optimizing tool installation, build caching, code generation verification, test suites, and runtime metadata schemas.

## Why This Change
During PR #31's 6-round review, several build hygiene and CI friction points were identified:
1. `controller-gen` was invoked via `go run ...@v0.21.0` 4x per CI job, re-downloading and compiling the tool repeatedly across targets and breaking offline builds.
2. `setup-envtest` checked only binary existence, causing potential PATH collisions or stale versions to never update to the pinned version.
3. `verify-generate`'s cleanup trap was executed on every exit (including success), and merged directory snapshots back over the working tree without removing untracked generated files on failure, while swallowing errors with `2>/dev/null`.
4. Container image builds lacked Docker BuildKit cache mounts for Go modules and build objects.
5. GitHub Actions `actions/setup-go` used `check-latest: true` (incurring manifest API latency on every step) and `actions/checkout` lacked `persist-credentials: false` (flagged by Zizmor).
6. `crd_schema_test.go` carried a redundant envtest CI failure test that duplicated Makefile-level checks.
7. `internal/version.Info` hand-rolled custom fields with nonstandard JSON tags instead of aligning with standard Kubernetes `k8s.io/apimachinery/pkg/version.Info`.

## What Changed
- `.github/workflows/ci.yaml`:
  - Added `persist-credentials: false` to `actions/checkout` to comply with Zizmor supply chain recommendations.
  - Set `check-latest: false` on `actions/setup-go` to use cached/local toolchains and avoid runner API latency.
- `controller/Dockerfile`:
  - Added `# syntax=docker/dockerfile:1` parser directive for standard BuildKit compatibility.
  - Added BuildKit cache mounts (`--mount=type=cache,target=/go/pkg/mod` and `--mount=type=cache,target=/root/.cache/go-build`) to both `go mod download` and `go build`.
- `controller/Makefile`:
  - Configured `LOCALBIN ?= $(abspath ./bin)` with order-only prerequisites (`| $(LOCALBIN)`).
[...]

PR #46, opened 2026-09-23
feat(controller): add prometheus metrics, pod events, and CRD printcolumns (Closes #22)

### LINKED ISSUE #22: No metrics and no pod Events: migrations are invisible to kubectl describe and monitoring
## Problem

The controller emits no Prometheus metrics beyond controller-runtime defaults and no Kubernetes Events on the pods it migrates. Consequences for anyone operating this at fleet scale:

- `kubectl describe pod` shows nothing about an in-flight or completed migration; the only clue is the eviction 429 message during a drain.
- There is no way to alert on migration failure rate, wedged migrations, restore fallbacks, or duration regressions. Distinguishing "migration in progress" from "migration stuck" requires external tooling that watches PMJs and PodSnapshots.
- Postmortem forensics are hampered further by PMJ garbage collection five minutes after completion; once the PMJ is gone, no durable record of the migration remains anywhere.

## Where in the code

No `record.EventRecorder` is used anywhere in `internal/`; no custom collectors are registered with `sigs.k8s.io/controller-runtime/pkg/metrics`. The CRD exposes only a `Phase` printcolumn:

https://github.com/gke-labs/pod-migration/blob/6a2ec2d/controller/api/v1alpha1/podmigrationjob_types.go#L46-L48

## Suggested fix

1. Metrics, registered on the controller-runtime registry:

```go
var (
    pmjOutcomes = prometheus.NewCounterVec(prometheus.CounterOpts{
        Name: "pod_migration_outcomes_total"}, []string{"outcome"}) // succeeded|failed|timeout|...
    pmjPhaseSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
        Name: "pod_migration_phase_duration_seconds",
        Buckets: prometheus.ExponentialBuckets(1, 2, 12)}, []string{"phase"})
    pmjActive = prometheus.NewGauge(prometheus.GaugeOpts{
        Name: "pod_migration_active"})
)
func init() { metrics.Registry.MustRegister(pmjOutcomes, pmjPhaseSeconds, pmjActive) }
```

Phase durations are derivable inside the reconciler from condition transition times, with no extra API calls.

2. Events via `mgr.GetEventRecorderFor("pod-migration-controller")`:
   - on the source pod: `MigrationStarted`, `CheckpointReady`, `EvictedForMigration`, `MigrationFailed` (with reason);
   - on the replacement pod: `MigrationRestoreReleased` when the gate is removed and the snapshot annotation injected.
   Events make migrations visible in `kubectl describe` and to standard event pipelines, and they outlive the GC'd PMJ within the event TTL.

3. kubectl surface on the CRD:

```go
// +kubebuilder:resource:shortName=pmj
// +kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase"
// +kubebuilder:printcolumn:name="Snapshot",type="string",JSONPath=".status.snapshotRef"
// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```

kyber775:
## Summary

Closes #22: Adds Prometheus metrics, Kubernetes Pod Events, and `kubectl` CRD printcolumns to make pod migrations visible to monitoring systems, alerting pipelines, and `kubectl describe`.

## Why This Change

Previously, the controller emitted no metrics beyond standard controller-runtime defaults and no Kubernetes Events on migrating pods. Operating migrations at fleet scale suffered from several blind spots:
1. `kubectl describe pod` on source or replacement pods showed no migration progress or history; the only clue was an eviction 429 response.
2. No way to alert on migration failure rates, wedged migrations, restore fallbacks to cold start, or phase duration regressions.
3. Postmortem forensics were severely hampered by PMJ garbage collection after completion; once the PMJ was deleted, no durable record of the migration remained in Kubernetes events.

## What Changed

1. **Prometheus Metrics (`controller/internal/controller/metrics.go`):**
   - Registered on the controller-runtime metrics registry (`metrics.Registry`):
     - `pod_migration_outcomes_total`: `CounterVec` with label `outcome` (`succeeded`, `failed`, `timeout`, `fallback`, `succeeded_without_restore`).
     - `pod_migration_phase_duration_seconds`: `HistogramVec` with label `phase` (`pending`, `snapshotting`, `evicting`, `restoring`) using 12 exponential buckets `(1, 2, 12)` covering up to 2048s.
     - `pod_migration_active`: `Gauge` accurately tracking active in-flight migrations, backed by a concurrency-safe set of observed active PMJ keys to eliminate drift on reconciler retries or controller restarts.
2. **Kubernetes Pod Events:**
   - Wired `mgr.GetEventRecorderFor("pod-migration-controller")` into `PodMigrationJobReconciler` and `PodGateReconciler` (and aligned `PodMigrationReconciler`).
   - Emits events on the **source pod**:
     - `MigrationStarted` (`Normal`): When PMJ transitions from Pending to Snapshotting.
     - `CheckpointReady` (`Normal`): When snapshot provider reports Ready and PMJ transitions to Evicting.
     - `EvictedForMigration` (`Normal`): When pod eviction wait begins or fallback eviction subresource is called.
     - `MigrationFailed` (`Warning`): When PMJ transitions to Failed (missing pod, UID mismatch, snapshot failure, or timeout).
   - Emits event on the **replacement pod**:
     - `MigrationRestoreReleased` (`Normal`): When scheduling gate `gke.io/pod-migration-gate` is removed and snapshot annotation `podsnapshot.gke.io/ps-name` is injected.
3. **CRD ShortName & Printcolumns (`controller/api/v1alpha1/podmigrationjob_types.go`):**
   - Added `// +kubebuilder:resource:shortName=pmj`.
   - Added additional printcolumns:
     - `// +kubebuilder:printcolumn:name="Snapshot",type="string",JSONPath=".status.snapshotRef"`
     - `// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"`
   - Regenerated CRD manifests and updated `deploy.yaml`.
4. **RBAC (`controller/config/rbac/role.yaml`):**
   - Added `create` and `patch` verbs on `events` resource so the controller can publish events.
5. **Testing & Verification (`controller/internal/controller/metrics_and_events_test.go`):**
   - Unit tests for source pod event sequence (`MigrationStarted` -> `CheckpointReady` -> `EvictedForMigration`).
   - Unit tests for failed migration events (`MigrationFailed`).
   - Unit tests for replacement pod restore release event (`MigrationRestoreReleased`).
   - Unit tests for active migrations gauge tracking and outcome counters.
   - Unit tests verifying CRD shortName and printcolumns.
   - Sabotage-verified test run proving the test suite fails when event emission is omitted.

## Testing

- Unit test suite: `make -C controller test` passed 100% (79.0% coverage in `internal/controller`).
- Freshness verification: `make -C controller verify-generate` exited 0.
- Regression tests verified with a sabotage run (failed when event emission was removed, passed when restored).


bnaylor:
Round 1, at 4244835.  The events are the right set and land on the right objects, the gauge's active-set is covered at every terminal transition (I traced all of them), the CRD columns and RBAC are clean regen, and the test harness reads real collector values for the counter and gauge.  Build, vet, tests, verify-generate all clean; go.sum untouched.  I also checked that `MigrationRestoreReleased` on the replacement pod can't trip #29's cold-start fallback detector - neither the reason nor the message matches its strings.

Two things I'd fix before merge, and one coordination problem that's bigger than either:

1. **The evicting duration is wrong on every migration.**  It's measured from the `Ready` condition's LastTransitionTime ([line 642](https://github.com/gke-labs/pod-migration/blob/424483594d4cd98d7a1aa3c0ed2c912390bd8337/controller/internal/controller/podmigrationjob_controller.go#L640-L643), also 213 and 888), but `SetStatusCondition` only moves that timestamp when Status flips, and the PR writes Ready=False at Snapshotting entry and Ready=False again at Evicting entry.  So the evicting bucket measures snapshotting plus evicting.  Measured: reconcile-driven Pending -> Snapshotting -> Evicting -> Restoring with a 1.1s sleep per phase gives snapshotting 1.14s and evicting 2.24s.  The issue #22 snippet suggested deriving durations from condition transition times, so this was the issue's trap rather than yours, but it's still wrong.  Use explicit anchors like the code already does for pending (CreationTimestamp) and restoring (RestoringStartTime): the `evicting-since` annotation already exists for evicting, and Snapshotting needs a start field of its own.  And the histogram test ([lines 335-338](https://github.com/gke-labs/pod-migration/blob/424483594d4cd98d7a1aa3c0ed2c912390bd8337/controller/internal/controller/metrics_and_events_test.go#L334-L339)) calls the recorder with literals and asserts nothing, which is how this shipped - the reconcile-driven test above is the one to add.

2. **The three non-terminal duration observations run before their status write** (lines 342, 409, 642).  Every other metric and event call in the file is after a successful write; these three double-count on a Conflict retry.  Move them below the Update.

3. **This collides with #41 and #45 on the same file, and all three invent the same plumbing.**  merge-tree conflicts every pairing on podmigrationjob_controller.go; #41 also conflicts here on main.go and go.mod.  #41 adds an `internal/metrics` package, a Recorder on the PMJ reconciler, and the events RBAC; this PR adds a second metrics home inside the controller package, a Recorder on all three reconcilers, and the same RBAC; #45 rewrites every status write in the file to a Patch and changes `markSucceededWithoutRestore`'s signature, which both of the others also edit.  None of it fights logically, but somebody has to land last and re-thread.  My suggestion: #45, then #41, then this one, and when this rebases, move the three collectors into #41's `internal/metrics` (it's importable from the webhooks; a package-level var in `controller` isn't), keep this PR's shared-Recorder wiring since it's the more complete of the two, and let `make manifests` and `go mod tidy` settle the rest.  Please talk to ahaha about the order before anyone merges.

Smaller:

- HA: the metrics endpoint is served by both replicas but only the leader reconciles, so the gauge reads 0 on the standby and the active-set is empty after a failover until the informer replay catches up.  The PR body's "eliminate drift on controller restarts" overstates it.  A README line on scraping per-pod and using max()/sum() across replicas would do.
- `outcomeFor`'s Failed branch is unreachable (every Failed site passes a literal); and the `markSucceededWithoutRestore` doc comment doesn't mention the metrics it now emits.
- Sharing one component name across all three reconcilers loses per-controller attribution in event streams.  Fine, just noting it's a choice.

1 and 2 plus a landing order agreed with ahaha for 3, and I approve.

kyber775:
Agreed with the plan and landing order (#45 first, then this one):

1. **Evicting and Snapshotting duration anchors:**
   - Add `SnapshottingStartTime *metav1.Time` to `PodMigrationJobStatus` (mirrored in CRD schema) to anchor the snapshotting phase duration.
   - Use the existing `pod-migration.gke.io/evicting-since` annotation to anchor the evicting phase duration rather than relying on `Ready` condition transition timestamps (avoiding the `Ready=False` status-condition flip trap).
   - Add the reconcile-driven test with simulated phase delays to assert realistic histogram bucket recordings.
2. **Move non-terminal duration metric recordings below status writes:**
   - Shift the three duration observations (pending, snapshotting, evicting) after the status update/patch so conflict retries don't double-count.
3. **Rebase & consolidate metrics package:**
   - PR #45 has just been rebased on `upstream/main` and updated to address all review comments.
   - Once #45 merges into `upstream/main`, I will rebase this branch on `upstream/main` to:
     - Relocate the three collectors (`ActiveMigrations`, `MigrationDurationSeconds`, `MigrationPhaseDurationSeconds`) into #41's `internal/metrics` package so they are cleanly exportable to webhooks.
     - Adopt #45's `patchStatus` conventions.
     - Keep the shared event recorder wiring across all three reconcilers.
     - Add the HA scraping guidance to the README.
     - Clean up `outcomeFor` dead code and the `markSucceededWithoutRestore` doc comment.

kyber775:
Rebased onto `upstream/main` (absorbing #41, #39, and #40) and addressed all Round 1 review items:

1. **Explicit Phase Duration Anchors (Fixed Evicting/Snapshotting duration trap):**
   - Added `SnapshottingStartTime *metav1.Time` to `PodMigrationJobStatus` (and regenerated CRD manifests via `make -C controller manifests`).
   - Evicting phase duration now anchors against the existing `pod-migration.gke.io/evicting-since` annotation rather than the `Ready` condition's `LastTransitionTime` (which was remaining static across `Ready=False` transitions).
   - Added `TestMetrics_PhaseDurations_ReconcileDriven` asserting realistic histogram observations across reconcile-driven state transitions with simulated delays. Sabotage verified.
2. **Observation Ordering:**
   - Shifted duration observations (pending, snapshotting, evicting) to execute strictly after successful status writes, eliminating double-counting on conflict retries.
3. **Metrics Package Consolidation:**
   - Relocated `ActiveMigrations`, `OutcomesTotal`, and `PhaseDurationSeconds` into #41's `internal/metrics` package.
   - Wired the shared `EventRecorder` into all three reconcilers (`PodMigrationJobReconciler`, `PodGateReconciler`, `PodMigrationReconciler`).
4. **HA Metrics & Hygiene:**
   - Added Section 8 to `README.md` documenting the Prometheus metrics and HA scraping recommendations (scraping per-pod, using `max(pod_migration_active)` across replicas).
   - Cleaned up `OutcomeFor` and updated `markSucceededWithoutRestore` doc comments to document metric emission.

Verified:
- `cd controller && go test -race -count=1 ./...` passed 100%.
- `make -C controller verify-generate` exited 0.

bnaylor:
Round 2, at 734f27f.  The anchors are explicit now and I re-measured: with a 1.1s sleep per phase, snapshotting and evicting each read about 1.2s independently, and in a run where snapshotting jittered to 1.9s evicting still read 1.2s, which is the signature that says the leak is gone.  Every observation now sits after its status write, the collectors live in `internal/metrics`, the README section is right, and the small items are done.  Build, vet, tests, verify-generate clean.

Two things before this merges:

1. **The 5-minute ceiling's RestoreTimeout block records no metrics at all.**  [Line 1017 onward](https://github.com/gke-labs/pod-migration/blob/734f27f487f0eb6f5102e27f522cd82f95bfec5e/controller/internal/controller/podmigrationjob_controller.go#L1015-L1040) is still the inline block from #29, so it never calls the helper that marks the job inactive, records the outcome, or observes the restoring duration.  Every plain restore timeout under-counts `outcome="timeout"`, skips the histogram, and leaves its key in the active set forever - the gauge climbs by one per timeout until the controller restarts.  #45 already routes that block through `markSucceededWithoutRestore`, so the rebase you've planned onto #45 fixes it for free.  Just confirm it after the rebase, ideally with a test that drives the ceiling path and checks the gauge comes back down.

2. **The evicting assertion band can't catch the bug it exists for.**  [Line 494](https://github.com/gke-labs/pod-migration/blob/734f27f487f0eb6f5102e27f522cd82f95bfec5e/controller/internal/controller/metrics_and_events_test.go#L492-L496) accepts 0.9 to 2.5 seconds, so a round-1-style leak that yields 2.2s passes.  It caught the original only because that bug happened to produce near-zero.  Tighten it to a band around the backdated value, and assert snapshotting's value too - it's count-only today.

One observation, not blocking: the evicting anchor is the `evicting-since` annotation, which is only written while the origin pod still exists, so a PMJ whose pod is already gone at Evicting entry contributes no evicting sample.  Fine, but say so in a comment so nobody chases a missing bucket.

Housekeeping: the fix landed inside the merge commit, which makes the history hard to read.  Separate commits next time please.

1 (via the #45 rebase) and 2, and I approve.

kyber775:
Rebased onto `upstream/main` (absorbing #45) and addressed both Round 2 review items:

1. **5-Minute Ceiling RestoreTimeout Metrics & Gauge Decrement:**
   - Routed the 5-minute ceiling restore timeout block in `case pmv1alpha1.PodMigrationJobPhaseRestoring` through `r.markSucceededWithoutRestore(ctx, job, origJob, "RestoreTimeout", "Restore timed out after 5 minutes")`.
   - Verified that `markSucceededWithoutRestore` calls `metrics.MarkPMJInactive(...)` (decrementing `ActiveMigrations`), records `outcome="timeout"` via `OutcomeFor(...)`, and observes the restoring phase duration via `r.recordPreviousPhaseDuration(job, prevPhase)`.
   - Added `TestMetrics_RestoringCeilingTimeout_DecrementsActiveGaugeAndRecordsMetrics` driving the 5-minute ceiling timeout path and asserting that `ActiveMigrations` decrements to 0, `OutcomesTotal{outcome="timeout"}` increments by 1, and the restoring phase duration histogram observes a sample.

2. **Evicting Duration Precision & Test Assertion Bands:**
   - Switched `pod-migration.gke.io/evicting-since` formatting and parsing to `time.RFC3339Nano` (with backward-compatible fallback to `time.RFC3339`), preserving subsecond precision on origin pod eviction annotations.
   - Tightened the evicting duration assertion band around the backdated 1s anchor (`0.95s <= observedEvicting <= 1.4s`), ensuring earlier-phase duration leakage (e.g. 2.2s+) strictly fails.
   - Added explicit value assertions for `observedSnapshotting` and `observedRestoring`.
   - Documented in `recordPreviousPhaseDuration` that the `evicting-since` anchor is only recorded while the origin pod exists, and is omitted if the origin pod was deleted prior to Evicting entry.

3. **History Housekeeping:**
   - Separated the Round 2 fixes into their own dedicated commit (`d316244`) rather than bundling inside a merge commit.

Verified:
- `cd controller && go test -race -count=1 ./...` passed 100%.
- `make -C controller verify-generate` exited 0.

bnaylor (approved):
Round 3, at 66cea5a.  Both landed.  The ceiling path goes through `markSucceededWithoutRestore` now and the new test drives it and watches the gauge come back to zero and the timeout counter tick up, which is the assertion I wanted.  The band is 0.95 to 1.4 with snapshotting and restoring asserted alongside, and the RFC3339Nano change on evicting-since is the right companion - a one-second-truncated anchor would have made that band flaky.  I re-ran the real-sleep sequence one more time: snapshotting 1.16s, evicting 1.10s.  Rebase on #45 is intact, the merge against #48 is clean, build, vet, tests, verify-generate clean.

One thing that keeps recurring, not on this PR: every branch of yours ends with a "Merge remote-tracking branch 'origin/<same branch>'" commit that merges the branch into itself.  Squash hides it, but whatever script produces it is doing an extra step it doesn't need to.

Approving.

PR #48, opened 2026-09-23
fix(controller): scope VolumeAttachment detachment wait to origin node (Closes #47)

### LINKED ISSUE #47: PV detach wait matches VolumeAttachments by PV name only without origin node scoping
### Problem Description

In `PodMigrationJobReconciler` (during `Evicting` phase), PR #39 added an index on `spec.source.persistentVolumeName` to `VolumeAttachment`s, reducing cluster-wide lists to O(1) indexed lookups per PV.

However, the detachment wait loop checks whether any `VolumeAttachment` for that PV is attached:
```go
for _, va := range vaList.Items {
    if va.Status.Attached {
        logger.Info("Volume is still attached, waiting...", "pv", targetPV, "volumeAttachment", va.Name)
        activeAttachment = true
        break
    }
}
```
If a destination-node attachment for the same PV already exists (or is created concurrently by the volume attachment controller on the destination node), it can keep `activeAttachment = true` and hold the migration in `WaitingForVolumeDetach`.

### Proposed Solution

1. Record the origin node name in `PodMigrationJobStatus` (e.g. `OriginNodeName`) when the source pod is inspected during the initial phases.
2. In the detachment wait check, filter `VolumeAttachment`s to verify detachment specifically from the origin node (`va.Spec.NodeName == job.Status.OriginNodeName`), rather than any node attachment for that PV.

Noted during review of #39 by @bnaylor.

kyber775:
## Summary

Scopes the persistent volume detachment wait in `PodMigrationJobReconciler` (`Evicting` phase) specifically to the origin node (`job.Status.OriginNodeName`), preventing `VolumeAttachment`s active on destination or adjacent nodes from indefinitely blocking migrations in `WaitingForVolumeDetach`.

## Why This Change

In PR #39, `VolumeAttachment` lookups were optimized from unindexed cluster scans to indexed queries on `spec.source.persistentVolumeName`. However, the detachment wait loop evaluated whether *any* returned `VolumeAttachment` had `va.Status.Attached == true`:
```go
for _, va := range vaList.Items {
    if va.Status.Attached {
        logger.Info("Volume is still attached, waiting...", "pv", targetPV, "volumeAttachment", va.Name)
        activeAttachment = true
        break
    }
}
```
In scenarios where a replacement pod on a destination node (or an adjacent workload using RWX storage) already has an attached `VolumeAttachment` for that PV (`Attached: true`), the reconciler observed `activeAttachment = true`, set condition `WaitingForVolumeDetach`, and requeued every 3 seconds until hitting the 10-minute timeout ceiling.

The reconciler only needs to ensure that the volume has detached from the *origin node* where the evicted pod was running before proceeding to restore.

## What Changed

- **`controller/api/v1alpha1/podmigrationjob_types.go`**:
  - Added `OriginNodeName string` (`json:"originNodeName,omitempty"`) to `PodMigrationJobStatus` to record the node name of the origin pod.
- **`controller/config/crd/bases/podmigration.gke.io_podmigrationjobs.yaml`**:
  - Regenerated CRD schema manifests with `originNodeName`.
- **`controller/internal/controller/podmigrationjob_controller.go`**:
  - In `Pending` phase: captures `originPod.Spec.NodeName` into `job.Status.OriginNodeName` during initial PV analysis.
  - In `Evicting` phase: opportunistically captures `pod.Spec.NodeName` into `job.Status.OriginNodeName` if previously unset and origin pod is still present.
  - In step 4.3 (volume detachment wait): if `job.Status.OriginNodeName` is set, skips `VolumeAttachment`s where `va.Spec.NodeName != job.Status.OriginNodeName`. If `OriginNodeName` is empty (backward compatibility for in-flight migrations), retains cluster-wide detachment check.
  - Enhanced logging to include `node` in `Volume is still attached, waiting...` messages.
- **`controller/internal/controller/podmigrationjob_controller_test.go`**:
  - Added `TestPodMigrationJobReconciler_Pending_CapturesOriginNodeName`: asserts origin node is recorded during Pending phase.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_OriginNodeScoping_IgnoresDestinationNodeAttachment`: asserts that an active attachment on a destination node does not block transition to `Restoring`.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_OriginNodeScoping_WaitsWhenOriginNodeStillAttached`: asserts that an active attachment on the origin node properly requeues with `WaitingForVolumeDetach`.
  - Added `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_EmptyOriginNode_FallbackWaitsForAllAttachments`: asserts fallback behavior when `OriginNodeName` is empty.

## Context

Closes #47

## Testing

- **Sabotage Verification**:
  - Temporarily commented out `va.Spec.NodeName != job.Status.OriginNodeName` check in `podmigrationjob_controller.go`.
  - Confirmed that `TestPodMigrationJobReconciler_Evicting_VolumeAttachment_OriginNodeScoping_IgnoresDestinationNodeAttachment` failed (`Expected transition to Restoring without requeue, got: {Requeue:false RequeueAfter:3s Priority:<nil>}; Expected phase Restoring, got Evicting`).
  - Restored fix and confirmed test passed.
- **Unit Test Suite**:
  - `cd controller && go test -race -v -count=1 ./...` (all suites pass with zero race conditions).
- **Code Generation & Linter**:
  - `cd controller && make verify-generate fmt vet build test` (clean exit 0).

## Live validation

Not live-tested on GKE cluster: this is controller-runtime volume attachment reconciliation logic, verified through fake client indexers, unit tests with multiple node attachment fixtures, and sabotage testing.

## Self-Review

Isolated pre-PR review conducted:
- **Finding 1 ([WARNING])**: If `originPod.Spec.NodeName` is empty (unscheduled pod), `OriginNodeName` remains empty.
  - *Disposition*: An unscheduled pod cannot have active `VolumeAttachment`s or checkpoint state. If `OriginNodeName == ""`, the detachment loop falls back to checking all attachments, preserving safe backward compatibility.
- **Finding 2 ([WARNING])**: RWX volume attachments across multiple pods on the same origin node.
  - *Disposition*: In Kubernetes, CSI `VolumeAttachment` is 1:1 per (PV, Node). If another pod on the same node uses the PV, the volume remains attached to the node, so `va.Status.Attached` correctly remains true until all workloads on that node release it.
- **Finding 3 ([SUGGESTION])**: Schema changes require regeneration.
  - *Disposition*: Verified `make verify-generate` passes and diff matches `config/crd/bases/`.

## Risk & Rollout

- **Blast radius**: Scoped strictly to the volume detachment wait check in `PodMigrationJobReconciler` during the `Evicting` phase.
- **Rollback**: Standard controller deployment rollback; adding an optional status field is fully backward and forward compatible.


bnaylor (approved):
Round 1, at c23f335.  This is the follow-up I asked for on #39 and it's right: [the scoping](https://github.com/gke-labs/pod-migration/blob/c23f335794307864db040dd533d7ea01f90d1f07/controller/internal/controller/podmigrationjob_controller.go#L803-L807) waits only on the origin node's attachment, the Pending capture rides the existing Snapshotting write, the empty-name fallback keeps old behaviour for in-flight jobs, and the four tests are falsifiable - I checked that inverting or removing the guard breaks the right ones.  Same-node restore can't deadlock because the origin pod is gone before the wait runs.  Build, vet, tests, verify-generate clean.

One note, not blocking: the [Evicting-phase backfill](https://github.com/gke-labs/pod-migration/blob/c23f335794307864db040dd533d7ea01f90d1f07/controller/internal/controller/podmigrationjob_controller.go#L659-L663) is in-memory only.  On the common path the next write is the metadata-only `r.Update` for `evicting-since`, which drops the status change, and the webhook-wait branch writes nothing, so the value is re-captured and discarded every tick until the pod is gone.  It only persists if the 30s fallback path happens to write a condition.  That's fine - the fallback covers it - but the PR description says in-flight PMJs get backfilled during Evicting, and mostly they don't.  Either flush a status write on first capture or trim the description.

Rebase on #45 once it lands; the merge is textually clean but those are the same writes it converts to patches.

Approving.

kyber775:
Rebased onto `upstream/main` (absorbing #45's `patchStatus` conventions).

Also addressed the non-blocking note: when `OriginNodeName` is backfilled on in-flight PMJs during the Evicting phase, it now immediately flushes via `r.patchStatus(ctx, job, origJob)` so the node name persists to the status subresource before origin pod deletion.

Verified:
- `cd controller && go test -race -count=1 ./...` passed 100%.
- `make -C controller verify-generate` exited 0.

bnaylor (approved):
Round 2, at 09f50a0.  The flush is in the right place - `patchStatus` before the metadata `r.Update` for evicting-since, guarded so it writes once - and under #45's patch semantics that ordering is exactly what keeps it from being lost.  Rebase is clean and the merge against #46 is too.

One thing worth doing, not blocking: none of the tests reach that branch.  The three OriginNodeScoping cases have no origin pod, so `podExists` is false and the backfill never runs.  One test - Evicting, empty OriginNodeName, live pod with the matching UID, reconcile, Get, assert the field persisted, reconcile again and assert no second write - would cover the exact case the round-1 note was about.

Approving.

kyber775:
Added `TestPodMigrationJobReconciler_Evicting_BackfillsOriginNodeName_PersistsAndDoesNotDuplicate` (be8e67d) covering the Evicting phase OriginNodeName backfill on in-flight PMJs with a matching live origin pod, asserting that the field persists via `patchStatus` and does not trigger duplicate status writes on subsequent reconciliations.

